PR #3, opened 2015-06-24
Add an IAntiforgery interface and simplify API

rynowak:


src/Microsoft.AspNet.Antiforgery/DefaultAntiforgery.cs
```diff
@@ -15,7 +15,7 @@ namespace Microsoft.AspNet.Antiforgery
     /// Provides access to the anti-forgery system, which provides protection against
     /// Cross-site Request Forgery (XSRF, also called CSRF) attacks.
     /// </summary>
-    public class Antiforgery
+    public class DefaultAntiforgery : IAntiforgery
```
dougbu: Thought we were moving away from `DefaultBlah` names? That is, why not leave this as `AntiForgery`?

src/Microsoft.AspNet.Antiforgery/IAntiforgery.cs
```diff
@@ -0,0 +1,74 @@
...
+    /// <summary>
+    /// Provides access to the antiforgery system, which provides protection against
+    /// Cross-site Request Forgery (XSRF, also called CSRF) attacks.
+    /// </summary>
+    public interface IAntiforgery
+    {
+        /// <summary>
+        /// Generates an input field for an antiforgery token. 
+        /// </summary>
+        /// <param name="context">The <see cref="HttpContext"/> associated with the current call.</param>
```
dougbu: call? Suggest "current request".

src/Microsoft.AspNet.Antiforgery/DefaultAntiforgery.cs
rynowak: There's no been no decision that I'm aware of this. The Antiforgery code now follows a consistent pattern. If we need to change it later it will be easy.

src/Microsoft.AspNet.Antiforgery/IAntiforgery.cs
rynowak: sure

PR #9, opened 2015-08-06
Fix aspnet/Mvc#2749 - fail gracefully with non-form content

rynowak:
This change will report a more specific error when antiforgery is used
with non-form content than "invalid content type".

### DIFF
--- a/test/Microsoft.AspNet.Antiforgery.Test/DefaultAntiforgeryTokenStoreTest.cs
+++ b/test/Microsoft.AspNet.Antiforgery.Test/DefaultAntiforgeryTokenStoreTest.cs
@@ -187,10 +187,43 @@ namespace Microsoft.AspNet.Antiforgery
         }
 
         [Fact]
+        public async Task GetRequestTokens_NonFormContentType_Throws()
+        {
+            // Arrange
+            var httpContext = new DefaultHttpContext();
+            httpContext.Request.ContentType = "application/json";
+
+            // Will not be accessed
+            httpContext.Request.Form = null;
+            httpContext.Request.Cookies = new ReadableStringCollection(new Dictionary<string, string[]>()
+            {
+                { "cookie-name", new string[] { "cookie-value" } },
+            });
+
+            var options = new AntiforgeryOptions()
+            {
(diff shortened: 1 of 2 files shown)


test/Microsoft.AspNet.Antiforgery.Test/DefaultAntiforgeryTokenStoreTest.cs
```diff
@@ -187,10 +187,43 @@ public void GetCookieToken_CookieIsValid_ReturnsToken()
...
+            var tokenStore = new DefaultAntiforgeryTokenStore(
+                optionsAccessor: new TestOptionsManager(options),
+                tokenSerializer: null);
+
+            // Act
+            var exception = await Assert.ThrowsAsync<InvalidOperationException>(
+                async () => await tokenStore.GetRequestTokensAsync(httpContext));
+
+            // Assert         
+            Assert.Equal("The required antiforgery form field \"form-field-name\" is not present.", exception.Message);
```
rynowak: Seemed like reusing the existing error was ok here.

test/Microsoft.AspNet.Antiforgery.Test/DefaultAntiforgeryTokenStoreTest.cs
```diff
@@ -220,6 +253,7 @@ public void GetCookieToken_CookieIsValid_ReturnsToken()
         {
             // Arrange
             var httpContext = new DefaultHttpContext();
+            httpContext.Request.ContentType = "application/x-www-form-urlencoded";
```
rynowak: This isn't strictly needed because `HasFormContentType` checks for `Form != null` but it seemed good for clarity.

test/Microsoft.AspNet.Antiforgery.Test/DefaultAntiforgeryTokenStoreTest.cs
Eilon: Yeah certainly good enough for now. If we see this causes confusion we could make a new message.

Eilon:
:shipit:

PR #18, opened 2015-10-09
React to aspnet/Universe#290 fix

dougbu:


dougbu:
More normal reaction PR

PR #21, opened 2015-11-11
Rybrande/return cookie token

ryanbrandenburg:
Always return the cookie token, weather it's old or new.

Needed for my current MusicStore PR.

### DIFF
--- a/src/Microsoft.AspNet.Antiforgery/DefaultAntiforgery.cs
+++ b/src/Microsoft.AspNet.Antiforgery/DefaultAntiforgery.cs
@@ -67,7 +67,10 @@ namespace Microsoft.AspNet.Antiforgery
             CheckSSLConfig(context);
 
             var tokenSet = GetTokensInternal(context);
-            SaveCookieTokenAndHeader(context, tokenSet.CookieToken);
+            if (tokenSet.IsNewCookieToken)
+            {
+                SaveCookieTokenAndHeader(context, tokenSet.CookieToken);
+            }
             return Serialize(tokenSet);
         }
 
@@ -226,8 +229,9 @@ namespace Microsoft.AspNet.Antiforgery
             return new AntiforgeryTokenSetInternal()
             {
                 // Note : The new cookie would be null if the old cookie is valid.
-                CookieToken = newCookieToken,
-                FormToken = formToken
+                CookieToken = cookieToken,
+                FormToken = formToken,
+                IsNewCookieToken = newCookieToken != null
             };
         }
 
@@ -243,6 +247,8 @@ namespace Microsoft.AspNet.Antiforgery
             public AntiforgeryToken FormToken { get; set; }
 
             public AntiforgeryToken CookieToken { get; set; }
+
+            public bool IsNewCookieToken { get; set; }
         }
     }
 }
--- a/test/Microsoft.AspNet.Antiforgery.Test/DefaultAntiforgeryTest.cs
+++ b/test/Microsoft.AspNet.Antiforgery.Test/DefaultAntiforgeryTest.cs
@@ -313,8 +313,49 @@ namespace Microsoft.AspNet.Antiforgery
             var tokenset = antiforgery.GetTokens(context.HttpContext);
 
             // Assert
-            Assert.Null(tokenset.CookieToken);
+            Assert.Equal("serialized-old-cookie-token", tokenset.CookieToken);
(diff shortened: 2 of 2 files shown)


src/Microsoft.AspNet.Antiforgery/DefaultAntiforgery.cs
```diff
@@ -243,6 +247,8 @@ private class AntiforgeryTokenSetInternal
             public AntiforgeryToken FormToken { get; set; }
 
             public AntiforgeryToken CookieToken { get; set; }
+
+            public bool CookieTokenIsNew { get; set; }
```
rynowak: Nit: `IsNewCookieToken`

rynowak:
Can I get a summary of the problem solved by this? The code changes are simple and seem safe. What's the overall goal?

test/Microsoft.AspNet.Antiforgery.Test/DefaultAntiforgeryTest.cs
```diff
@@ -313,11 +313,30 @@ public void GetTokens_ExistingValidCookieToken_GeneratesANewFormToken()
...
+        {
+            var context = CreateMockContext(
+                new AntiforgeryOptions(),
+                useOldCookie: true,
+                isOldCookieValid: true);
+            var antiforgery = GetAntiforgery(context);
+
+            var tokenSet = antiforgery.GetAndStoreTokens(context.HttpContext);
+            
+            //Strict mock would have complained if SaveCookieToken was called.
```
rynowak: CS01092: Comment not useful at DefaultAntiforgeryTest.cs (332:12)

test/Microsoft.AspNet.Antiforgery.Test/DefaultAntiforgeryTest.cs
```diff
@@ -313,11 +313,30 @@ public void GetTokens_ExistingValidCookieToken_GeneratesANewFormToken()
             var tokenset = antiforgery.GetTokens(context.HttpContext);
 
             // Assert
-            Assert.Null(tokenset.CookieToken);
+            Assert.NotNull(tokenset.CookieToken);
```
rynowak: These `Assert.NotNulls` really aren't needed.

test/Microsoft.AspNet.Antiforgery.Test/DefaultAntiforgeryTest.cs
```diff
@@ -313,11 +313,30 @@ public void GetTokens_ExistingValidCookieToken_GeneratesANewFormToken()
...
+            var antiforgery = GetAntiforgery(context);
+
+            var tokenSet = antiforgery.GetAndStoreTokens(context.HttpContext);
+            
+            //Strict mock would have complained if SaveCookieToken was called.
+
+            Assert.NotNull(tokenSet.CookieToken);
+            Assert.Equal("serialized-old-cookie-token", tokenSet.CookieToken);
+            Assert.Equal("serialized-form-token", tokenSet.FormToken);
+        }
```
rynowak: Where are the unit tests verifying that fancy new property?

rynowak:
:watch:

Eilon:
@blowdart because @blowdart and because anti-forgery. Just making sure you're cool with this.

ryanbrandenburg:
@rynowak the problem being solved here is that if there was already a cookie token existing when you call GetTokens it doesn't give you a cookie token, only a new form token. Since calling GetTokens is apparently how we want people to do Anti-forgery for AJAX calls it is important that it give you both of these.

test/Microsoft.AspNet.Antiforgery.Test/DefaultAntiforgeryTest.cs
ryanbrandenburg: The above test was intended to be the test of the property. If the CookieToken is not new, then we shouldn't be trying to save it again. I've added a test to ensure we save it when we actually do create a new token as well, will put that up with the update.

ryanbrandenburg:
:up::date:

rynowak:
Groovy :shipit:

blowdart:
Damnit, I can't find anything to pick on. This sets a worrying precedent for new Ryan. :shipit:

Eilon:
@ryanbrandenburg FYI that was a huge compliment from @blowdart :smile:

PR #31, opened 2015-12-17
Add a sample demonstrating Antiforgery with AJAX

rynowak:


samples/AntiforgerySample/Startup.cs
```diff
@@ -10,13 +23,55 @@ public class Startup
...
+
+            services.AddSingleton<TodoRepository>();
         }
 
-        public void Configure(IApplicationBuilder app)
+        public void Configure(IApplicationBuilder app, IAntiforgery antiforgery, IOptions<AntiforgeryOptions> options, TodoRepository repository)
         {
+            app.Use(next => context =>
+            {
+                if (context.Request.Path == "/")
```
kichalla: probably add one where the path is "/index.html"?

samples/AntiforgerySample/Startup.cs
```diff
@@ -10,13 +23,55 @@ public class Startup
...
+                {
+                    // We can send the request token as a JavaScript-readable cookie, and Angular will use it by default.
+                    var tokens = antiforgery.GetAndStoreTokens(context);
+                    context.Response.Cookies.Append("XSRF-TOKEN", tokens.RequestToken, new CookieOptions() { HttpOnly = false });
```
kichalla: you probably need to set this name in options too?

rynowak: It would be nice to find a way to build this in as well. We could make this a standalone middleware if necessary. Right now it's up to the app developer to know when to set the cookie.

kichalla:
:shipit:

samples/AntiforgerySample/Startup.cs
rynowak: Sure

PR #35, opened 2016-01-11
Build with dotnet

victorhurdugaci:
cc @Tratcher @ryanbrandenburg

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,7 +1,7 @@
 init:
   - git config --global core.autocrlf true
 build_script:
-  - build.cmd --quiet --parallel verify
+  - build.cmd --quiet verify
 clone_depth: 1
 test: off
 deploy: off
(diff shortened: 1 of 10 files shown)


appveyor.yml
```diff
@@ -1,7 +1,7 @@
 init:
   - git config --global core.autocrlf true
 build_script:
-  - build.cmd --quiet --parallel verify
```
ryanbrandenburg: Is it that we don't want it parallel anymore or does that idea just not exist in the new world?

pranavkm: @ryanbrandenburg some of these come from the Universe template - https://github.com/aspnet/Universe/blob/dev/build-template-dotnet/appveyor.yml. If we find issues with these later, we'll bulk update all the repos. For now, I'd focus on code changes

test/Microsoft.AspNet.Antiforgery.FunctionalTests/project.json
```diff
@@ -2,23 +2,24 @@
...
   "commands": {
-    "run": "xunit.runner.aspnet",
     "test": "xunit.runner.aspnet"
   },
   "frameworks": {
     "dnx451": {
       "dependencies": {
         "Moq": "4.2.1312.1622",
-        "System.Net.Http": "4.0.1-rc2-23621"
+        "System.Net.Http": "4.0.1-rc2-23621",
```
ryanbrandenburg: I know it was that way before but is this hyper-specific for a reason?

victorhurdugaci: Not sure about that...

PR #39, opened 2016-02-02
Remove dependency on routing in sample

rynowak:
Removing the routing dependency since this is moving lower in the stack.


dougbu:
:shipit:

Eilon:
:shipit:

PR #40, opened 2016-02-02
Adds a IsRequestValidAsync method

rynowak:
We want a non-throwing way to validate for use in auth.

Some other misc cleanup
- docs for IAntiforgeryTokenGenerator
~~- Add HttpContext parameter where to all IAntiforgeryGenerator methods~~
- rename parameters on DefaultAntiforgery

### DIFF
--- a/src/Microsoft.AspNetCore.Antiforgery/DefaultAntiforgeryTokenGenerator.cs
+++ b/src/Microsoft.AspNetCore.Antiforgery/DefaultAntiforgeryTokenGenerator.cs
@@ -22,7 +22,7 @@ namespace Microsoft.AspNetCore.Antiforgery
         }
 
         /// <inheritdoc />
-        public AntiforgeryToken GenerateCookieToken(HttpContext httpContext)
+        public AntiforgeryToken GenerateCookieToken()
         {
             return new AntiforgeryToken()
             {
@@ -46,10 +46,10 @@ namespace Microsoft.AspNetCore.Antiforgery
                 throw new ArgumentNullException(nameof(cookieToken));
             }
 
-            if (!IsCookieTokenValid(httpContext, cookieToken))
+            if (!IsCookieTokenValid(cookieToken))
             {
                 throw new ArgumentException(
-                    Resources.Antiforgery_CookieToken_MustBeProvided_Generic,
+                    Resources.Antiforgery_CookieToken_IsInvalid,
                     nameof(cookieToken));
             }
 
@@ -99,7 +99,7 @@ namespace Microsoft.AspNetCore.Antiforgery
         }
 
         /// <inheritdoc />
-        public bool IsCookieTokenValid(HttpContext httpContext, AntiforgeryToken cookieToken)
+        public bool IsCookieTokenValid(AntiforgeryToken cookieToken)
         {
             return cookieToken != null && cookieToken.IsCookieToken;
         }
--- a/src/Microsoft.AspNetCore.Antiforgery/DefaultAntiforgeryTokenGenerator.cs
+++ b/src/Microsoft.AspNetCore.Antiforgery/DefaultAntiforgeryTokenGenerator.cs
@@ -2,7 +2,6 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
-using System.Diagnostics;
 using System.Security.Claims;
 using System.Security.Principal;
 using Microsoft.AspNetCore.Http;
@@ -22,7 +21,8 @@ namespace Microsoft.AspNetCore.Antiforgery
             _additionalDataProvider = additionalDataProvider;
         }
 
-        public AntiforgeryToken GenerateCookieToken()
+        /// <inheritdoc />
+        public AntiforgeryToken GenerateCookieToken(HttpContext httpContext)
         {
             return new AntiforgeryToken()
             {
@@ -31,6 +31,7 @@ namespace Microsoft.AspNetCore.Antiforgery
             };
         }
 
+        /// <inheritdoc />
         public AntiforgeryToken GenerateRequestToken(
             HttpContext httpContext,
             AntiforgeryToken cookieToken)
@@ -40,7 +41,17 @@ namespace Microsoft.AspNetCore.Antiforgery
                 throw new ArgumentNullException(nameof(httpContext));
             }
 
-            Debug.Assert(IsCookieTokenValid(cookieToken));
+            if (cookieToken == null)
+            {
+                throw new ArgumentNullException(nameof(cookieToken));
+            }
+
+            if (!IsCookieTokenValid(httpContext, cookieToken))
+            {
+                throw new ArgumentException(
+                    Resources.Antiforgery_CookieToken_MustBeProvided_Generic,
+                    nameof(cookieToken));
+            }
 
             var requestToken = new AntiforgeryToken()
             {
@@ -87,15 +98,18 @@ namespace Microsoft.AspNetCore.Antiforgery
             return requestToken;
         }
 
-        public bool IsCookieTokenValid(AntiforgeryToken cookieToken)
+        /// <inheritdoc />
+        public bool IsCookieTokenValid(HttpContext httpContext, AntiforgeryToken cookieToken)
         {
-            return (cookieToken != null && cookieToken.IsCookieToken);
+            return cookieToken != null && cookieToken.IsCookieToken;
         }
 
-        public void ValidateTokens(
+        /// <inheritdoc />
+        public bool TryValidateTokenSet(
             HttpContext httpContext,
             AntiforgeryToken cookieToken,
-            AntiforgeryToken requestToken)
+            AntiforgeryToken requestToken,
+            out string message)
         {
             if (httpContext == null)
             {
@@ -119,13 +133,15 @@ namespace Microsoft.AspNetCore.Antiforgery
             // Do the tokens have the correct format?
             if (!cookieToken.IsCookieToken || requestToken.IsCookieToken)
             {
-                throw new AntiforgeryValidationException(Resources.AntiforgeryToken_TokensSwapped);
+                message = Resources.AntiforgeryToken_TokensSwapped;
+                return false;
             }
 
             // Are the security tokens embedded in each incoming token identical?
             if (!object.Equals(cookieToken.SecurityToken, requestToken.SecurityToken))
             {
-                throw new AntiforgeryValidationException(Resources.AntiforgeryToken_SecurityTokenMismatch);
+                message = Resources.AntiforgeryToken_SecurityTokenMismatch;
+                return false;
             }
 
             // Is the incoming token meant for the current user?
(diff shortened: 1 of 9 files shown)


src/Microsoft.AspNetCore.Antiforgery/DefaultAntiforgeryTokenGenerator.cs
```diff
@@ -40,7 +41,17 @@ public AntiforgeryToken GenerateCookieToken()
                 throw new ArgumentNullException(nameof(httpContext));
             }
 
-            Debug.Assert(IsCookieTokenValid(cookieToken));
```
rynowak: This is a public method, this shouldn't be a debug.assert

src/Microsoft.AspNetCore.Antiforgery/DefaultAntiforgeryTokenGenerator.cs
```diff
@@ -22,7 +21,8 @@ public class DefaultAntiforgeryTokenGenerator : IAntiforgeryTokenGenerator
             _additionalDataProvider = additionalDataProvider;
         }
 
-        public AntiforgeryToken GenerateCookieToken()
+        /// <inheritdoc />
+        public AntiforgeryToken GenerateCookieToken(HttpContext httpContext)
```
dougbu: Why does this have a `HttpContext` parameter?

src/Microsoft.AspNetCore.Antiforgery/DefaultAntiforgeryTokenGenerator.cs
```diff
@@ -40,7 +41,17 @@ public AntiforgeryToken GenerateCookieToken()
...
-            Debug.Assert(IsCookieTokenValid(cookieToken));
+            if (cookieToken == null)
+            {
+                throw new ArgumentNullException(nameof(cookieToken));
+            }
+
+            if (!IsCookieTokenValid(httpContext, cookieToken))
+            {
+                throw new ArgumentException(
+                    Resources.Antiforgery_CookieToken_MustBeProvided_Generic,
```
dougbu: The message `The required antiforgery cookie token must be provided.` is misleading here. Since we know the caller passed a non-`null` `cookieToken`, the only possible problem is that the caller passed a non-cookie `AntiForgeryToken`. Seems more like `AntiforgeryToken_SecurityTokenMismatch` or `AntiforgeryToken_TokensSwapped`.

test/Microsoft.AspNetCore.Antiforgery.Test/DefaultAntiforgeryTest.cs
```diff
@@ -40,7 +40,7 @@ public class DefaultAntiforgeryTest
         }
 
         [Fact]
-        public void ChecksSSL_ValidateTokens_Throws()
+        public void ChecksSSL_TryValidateTokenSet_Throws()
```
dougbu: Name was better before. `TryValidateTokenSet` get's called but that method certainly isn't throwing.

src/Microsoft.AspNetCore.Antiforgery/DefaultAntiforgeryTokenGenerator.cs
rynowak: I'll just add a new message since none of these really seem like an exact fit.

`IsValidCookieToken` is extensible so, it's probably for the best.

src/Microsoft.AspNetCore.Antiforgery/DefaultAntiforgeryTokenGenerator.cs
```diff
@@ -87,15 +98,18 @@ public AntiforgeryToken GenerateCookieToken()
             return requestToken;
         }
 
-        public bool IsCookieTokenValid(AntiforgeryToken cookieToken)
+        /// <inheritdoc />
+        public bool IsCookieTokenValid(HttpContext httpContext, AntiforgeryToken cookieToken)
```
dougbu: Why does this method have a `HttpContext` parameter?

test/Microsoft.AspNetCore.Antiforgery.Test/DefaultAntiforgeryTest.cs
```diff
@@ -393,11 +386,13 @@ public void ValidateTokens_FromInvalidStrings_Throws()
...
+            string message;
             context.TokenGenerator
-                .Setup(o => o.ValidateTokens(
+                .Setup(o => o.TryValidateTokenSet(
                     context.HttpContext,
                     context.TestTokenSet.OldCookieToken,
-                    context.TestTokenSet.RequestToken))
+                    context.TestTokenSet.RequestToken,
+                    out message))
                 .Throws(new InvalidOperationException("my-message"));
```
dougbu: Why is `TryValidateTokenSet` set up to `throw` instead of returning `false`?

test/Microsoft.AspNetCore.Antiforgery.Test/DefaultAntiforgeryTest.cs
```diff
@@ -381,7 +374,7 @@ public void GetAndStoreTokens_NoExistingCookieToken_Saved()
         }
 
         [Fact]
-        public void ValidateTokens_FromInvalidStrings_Throws()
+        public void TryValidateTokenSet_FromInvalidStrings_Throws()
```
dougbu: More odd naming since this method isn't what's throwing nor what's called in `// Act`.

test/Microsoft.AspNetCore.Antiforgery.Test/DefaultAntiforgeryTest.cs
```diff
@@ -459,16 +456,67 @@ public void ValidateTokens_MissingCookieInTokenSet_Throws()
...
+            var context = CreateMockContext(new AntiforgeryOptions());
+
+            string message;
+            context.TokenGenerator
+                .Setup(o => o.TryValidateTokenSet(
+                    context.HttpContext,
+                    context.TestTokenSet.OldCookieToken,
+                    context.TestTokenSet.RequestToken,
+                    out message))
+                .Throws(new InvalidOperationException("my-message"));
```
dougbu: Return `false`.

test/Microsoft.AspNetCore.Antiforgery.Test/DefaultAntiforgeryTest.cs
```diff
@@ -459,16 +456,67 @@ public void ValidateTokens_MissingCookieInTokenSet_Throws()
...
+            var context = CreateMockContext(new AntiforgeryOptions());
+
+            string message;
+            context.TokenGenerator
+                .Setup(o => o.TryValidateTokenSet(
+                    context.HttpContext,
+                    context.TestTokenSet.OldCookieToken,
+                    context.TestTokenSet.RequestToken,
+                    out message))
+                .Verifiable();
```
dougbu: Return `true`.

test/Microsoft.AspNetCore.Antiforgery.Test/DefaultAntiforgeryTest.cs
```diff
@@ -423,11 +418,13 @@ public void ValidateTokens_FromValidStrings_TokensValidatedSuccessfully()
...
+            string message;
             context.TokenGenerator
-                .Setup(o => o.ValidateTokens(
+                .Setup(o => o.TryValidateTokenSet(
                     context.HttpContext,
                     context.TestTokenSet.OldCookieToken,
-                    context.TestTokenSet.RequestToken))
+                    context.TestTokenSet.RequestToken,
+                    out message))
                 .Verifiable();
```
dougbu: `.Returns(true)`

src/Microsoft.AspNetCore.Antiforgery/DefaultAntiforgeryTokenGenerator.cs
dougbu: :+1:

dougbu:
:watch:

src/Microsoft.AspNetCore.Antiforgery/DefaultAntiforgeryTokenGenerator.cs
rynowak: Any implementation other than ours would likely need it. Our default implementation just generates a 'dump' token. If you wanted to do something based on the user's claims/identity you'd need the context.

Basically it seemed very short-sighted and limiting that the methods operating on the request token have access to context, and the methods operating on the cookie token do not.

dougbu: > short-sighted and limiting

Hmm, also looks :100:% intentional. See for example the different way an `AntiforgeryToken` is serialized if `IsCookieToken` is `true`. That code "knows" there's no useful user-related information in a cookie token and doesn't attempt to serialize it.

rynowak: That still doesn't preclude another implementation.

test/Microsoft.AspNetCore.Antiforgery.Test/DefaultAntiforgeryTest.cs
rynowak: Yeah, bad find+replace it looks like

rynowak:
@dougbu updated

src/Microsoft.AspNetCore.Antiforgery/DefaultAntiforgeryTokenGenerator.cs
rynowak: Discussed more and agreed not to do this - we're not convinced that this change would do enough. We'll take a fresh look at this API if anyone asks for this kind of extensibility.

test/Microsoft.AspNetCore.Antiforgery.Test/DefaultAntiforgeryTest.cs
```diff
@@ -40,6 +41,28 @@ public class DefaultAntiforgeryTest
...
+
+            var antiforgery = GetAntiforgery(options);
+
+            // Act & Assert
+            var exception = await Assert.ThrowsAsync<InvalidOperationException>(
+                        async () => await antiforgery.IsRequestValidAsync(httpContext));
+            Assert.Equal(
+             @"The antiforgery system has the configuration value AntiforgeryOptions.RequireSsl = true, " +
+             "but the current request is not an SSL request.",
+             exception.Message);
```
dougbu: Moving a commit comment: Moar indent

test/Microsoft.AspNetCore.Antiforgery.Test/DefaultAntiforgeryTest.cs
```diff
@@ -393,17 +419,20 @@ public void ValidateTokens_FromInvalidStrings_Throws()
...
 
-            context.TokenGenerator
-                .Setup(o => o.ValidateTokens(
-                    context.HttpContext,
-                    context.TestTokenSet.OldCookieToken,
-                    context.TestTokenSet.RequestToken))
-                .Throws(new InvalidOperationException("my-message"));
-            context.TokenStore = null;
-            var antiforgery = GetAntiforgery(context);
+            // You can't really do Moq with out-parameters :(
```
dougbu: Moving a commit comment: You can but it's a pain. Not worth the contortions.

src/Microsoft.AspNetCore.Antiforgery/IAntiforgeryTokenGenerator.cs
```diff
@@ -10,23 +10,44 @@ namespace Microsoft.AspNetCore.Antiforgery
...
     {
-        // Generates a new random cookie token.
+        /// <summary>
+        /// Generates a new random cookie token.
+        /// </summary>
+        /// <returns>An <see cref="AntiforgeryToken"/>.</returns>
         AntiforgeryToken GenerateCookieToken();
 
         // Given a cookie token, generates a corresponding request token.
         // The incoming cookie token must be valid.
```
dougbu: Minor but these comments don't add anything over or above the new doc comments.

rynowak: OOps, I intended to delete this

test/Microsoft.AspNetCore.Antiforgery.Test/DefaultAntiforgeryTest.cs
```diff
@@ -40,6 +41,28 @@ public class DefaultAntiforgeryTest
...
+            var options = new AntiforgeryOptions()
+            {
+                RequireSsl = true
+            };
+
+            var antiforgery = GetAntiforgery(options);
+
+            // Act & Assert
+            var exception = await Assert.ThrowsAsync<InvalidOperationException>(
+                        async () => await antiforgery.IsRequestValidAsync(httpContext));
```
dougbu: Moving a commit comment: Can't this be done w/o `async` or `await` on this line? (Yah know, unnecessary boilerplate.)

dougbu:
:shipit:

PR #42, opened 2016-02-03
Add form and header name to token set

rynowak:


src/Microsoft.AspNetCore.Antiforgery/AntiforgeryTokenSet.cs
```diff
@@ -15,28 +15,57 @@ public class AntiforgeryTokenSet
         /// </summary>
         /// <param name="requestToken">The token that is supplied in the request.</param>
         /// <param name="cookieToken">The token that is supplied in the request cookie.</param>
-        public AntiforgeryTokenSet(string requestToken, string cookieToken)
+        /// <param name="formFieldName">The name of the form field used for the request token.</param>
+        /// <param name="headerName">The name of the header used for the request token.</param>
```
dougbu: `cookie token`

dougbu: Hmm, may be clearer to use `<paramref name="requestToken"/>` and `<paramref name="cookieToken"/>`.

dougbu: Talked offline and I didn't grok the system correctly. Both new parameters should mention `<paramref name="requestToken"/>`.

rynowak: I don't agree. These aren't at all related to the parameter, or code you write. They refer to concepts in the system as a whole.

src/Microsoft.AspNetCore.Antiforgery/AntiforgeryTokenSet.cs
```diff
@@ -15,28 +15,57 @@ public class AntiforgeryTokenSet
...
             {
                 throw new ArgumentException(Resources.ArgumentCannotBeNullOrEmpty, nameof(requestToken));
             }
 
+            if (string.IsNullOrEmpty(formFieldName))
+            {
+                throw new ArgumentException(Resources.ArgumentCannotBeNullOrEmpty, nameof(formFieldName));
+            }
+
+            if (string.IsNullOrEmpty(headerName))
```
dougbu: User can set `AntiforgeryOptions.HeaderName` to `null` or `string.Empty` w/o error. Further `null` is special cased later to mean "don't look for a header". So parameter validation is incorrect here.

src/Microsoft.AspNetCore.Antiforgery/AntiforgeryTokenSet.cs
```diff
@@ -15,28 +15,57 @@ public class AntiforgeryTokenSet
...
+            string cookieToken,
+            string formFieldName,
+            string headerName)
         {
             if (string.IsNullOrEmpty(requestToken))
             {
                 throw new ArgumentException(Resources.ArgumentCannotBeNullOrEmpty, nameof(requestToken));
             }
 
+            if (string.IsNullOrEmpty(formFieldName))
```
dougbu: I don't see anywhere else that `string.Empty` for the form field name causes an issue. Perfectly fine to gen a hidden field with that name.

dougbu:
:shipit: w/ fixed parameter validations.

PR #43, opened 2016-02-03
Move Antiforgery implementation details to `.Internal`

dougbu:
- move the files
- update the namespaces


dougbu:
/cc @rynowak
FYI @Eilon

dougbu:
Recommend looking at the namespace changes and checking out what's left in
- https://github.com/aspnet/Antiforgery/tree/dougbu/be.internal/src/Microsoft.AspNetCore.Antiforgery
- https://github.com/aspnet/Antiforgery/tree/dougbu/be.internal/test/Microsoft.AspNetCore.Antiforgery.Test

src/Microsoft.AspNetCore.Antiforgery/Internal/IAntiforgeryAdditionalDataProvider.cs
```diff
@@ -3,7 +3,7 @@
 
 using Microsoft.AspNetCore.Http;
 
-namespace Microsoft.AspNetCore.Antiforgery
+namespace Microsoft.AspNetCore.Antiforgery.Internal
```
rynowak: Do we want to leave this one public? This is the one documented extensibility point that I'm aware of: http://www.asp.net/mvc/overview/security/xsrfcsrf-prevention-in-aspnet-mvc-and-web-pages#_Scenarios_with_special

rynowak:
:shipit: after resolving what to do with the additonal data provider

src/Microsoft.AspNetCore.Antiforgery/Internal/IAntiforgeryAdditionalDataProvider.cs
dougbu: :ok: forgot about that case. Will move it back to top-level namespace.

dougbu:
705c080

PR #46, opened 2016-02-04
Move exceptions from the store to the facade

rynowak:
My earlier change to add TryValidateRequestAsync didn't go far enough,
because the store will still throw when the tokens aren't present. This
change is to make the store just return null tokens in these cases, and
move the exceptions to DefaultAntiforgery.

### DIFF
--- a/src/Microsoft.AspNetCore.Antiforgery/AntiforgeryTokenSet.cs
+++ b/src/Microsoft.AspNetCore.Antiforgery/AntiforgeryTokenSet.cs
@@ -23,22 +23,13 @@ namespace Microsoft.AspNetCore.Antiforgery
             string formFieldName,
             string headerName)
         {
-            if (string.IsNullOrEmpty(requestToken))
-            {
-                throw new ArgumentException(Resources.ArgumentCannotBeNullOrEmpty, nameof(requestToken));
-            }
-
             if (formFieldName == null)
             {
                 throw new ArgumentNullException(nameof(formFieldName));
             }
 
             RequestToken = requestToken;
-
-            // Cookie Token is allowed to be null in the case when the old cookie is valid
-            // and there is no new cookieToken generated.
             CookieToken = cookieToken;
-
             FormFieldName = formFieldName;
             HeaderName = headerName;
         }
--- a/src/Microsoft.AspNetCore.Antiforgery/Internal/DefaultAntiforgery.cs
+++ b/src/Microsoft.AspNetCore.Antiforgery/Internal/DefaultAntiforgery.cs
@@ -93,6 +93,10 @@ namespace Microsoft.AspNetCore.Antiforgery.Internal
             CheckSSLConfig(httpContext);
 
             var tokens = await _tokenStore.GetRequestTokensAsync(httpContext);
+            if (tokens.CookieToken == null || tokens.RequestToken == null)
+            {
+                return false;
+            }
 
             // Extract cookie & request tokens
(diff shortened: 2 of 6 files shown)


rynowak:
@dougbu - rebased and tests fixed for the new behavior

src/Microsoft.AspNetCore.Antiforgery/Internal/DefaultAntiforgery.cs
```diff
@@ -118,6 +122,33 @@ public AntiforgeryTokenSet GetTokens(HttpContext httpContext)
...
+                    var message = Resources.FormatAntiforgery_HeaderToken_MustBeProvided(_options.HeaderName);
+                    throw new AntiforgeryValidationException(message);
+                }
+                else
+                {
+                    var message = Resources.FormatAntiforgery_RequestToken_MustBeProvided(
+                        _options.FormFieldName,
+                        _options.HeaderName);
+                    throw new AntiforgeryValidationException(message);
+                }
```
dougbu: Not mandatory but this would be more readable w/o the `else`s and unburying the final block. Only downside is need to rename the last `message`.

src/Microsoft.AspNetCore.Antiforgery/AntiforgeryTokenSet.cs
```diff
@@ -23,22 +23,13 @@ public class AntiforgeryTokenSet
             string formFieldName,
             string headerName)
         {
-            if (string.IsNullOrEmpty(requestToken))
-            {
-                throw new ArgumentException(Resources.ArgumentCannotBeNullOrEmpty, nameof(requestToken));
-            }
```
dougbu: Previously we had an invariant that the `AntiforgeryTokenSet` could not be invalid. Probably :ok: as long as `DefaultAntiforgery` can never return such a token set. But have you confirmed we're covered?

test/Microsoft.AspNetCore.Antiforgery.Test/Internal/DefaultAntiforgeryTokenStoreTest.cs
```diff
@@ -145,7 +145,7 @@ public void GetCookieToken_CookieIsValid_ReturnsToken()
         }
 
         [Fact]
-        public async Task GetRequestTokens_CookieIsEmpty_Throws()
+        public async Task GetRequestTokens_CookieIsEmpty_ReturnsNullCookie()
```
dougbu: `...NullTokens` (plural)

dougbu:
:shipit: with small fixes

src/Microsoft.AspNetCore.Antiforgery/AntiforgeryTokenSet.cs
rynowak: I changed `DefaultAntiforgery.Serialize` to prove that it will never return a null token.

test/Microsoft.AspNetCore.Antiforgery.Test/Internal/DefaultAntiforgeryTokenStoreTest.cs
rynowak: Sure

src/Microsoft.AspNetCore.Antiforgery/Internal/DefaultAntiforgery.cs
rynowak: Don't agree :see_no_evil:

PR #48, opened 2016-02-04
Remove GetHtml

rynowak:
It's being moved to MVC


dougbu:
:shipit: if or once the MVC change is in.

Issue #30, opened 2015-12-16
Review generation of UID in DefaultClaimUidExtractor


rynowak:
According to @blowdart and @lodejard the way we generate the UID for a `ClaimsPrincipal` is wrong in antiforgery.

We should use the `ClaimsIdentity.NameClaimType` of **all** authenticated identities to generate the UID rather than looking for a claim of type `ClaimTypes.NameIdentifier`.

We need to also figure out the right thing to do (or if it's possible) when the user does not have a name.

Eilon:
@ajaybhargavb please discuss with @Tratcher to understand the multi-name and no-name scenarios.

brockallen:
Side note related to this: anti-forgery tokens that are issued when a user is anonymous, and then the user becomes authenticated caused problems in the past. This was due to the inclusion of the user's id in the xsrf token. Will this still be the case in asp.net 5?

leastprivilege:
The `Name` claims is a display name - you want to use a claim that has the unique id of the user - that would be the `sub` claim (https://openid.net/specs/openid-connect-core-1_0.html#StandardClaims) - but since you are stuck in medieval aka WS* times - `NameIdentifier` comes closest.

Historically `Name` or `IIdentity.Name` had no clear semantics and was often used for one or the other - the MS templates used it to display the "unique" name - e.g. "Hello Domain\User" (which is perfectly natural language ;))

IOW - the Name claim is not the right - the templates should set both nameid and Name - and use NameId for anti-forgery to educate people. If NameId is not set - thrown an exception.

Now give me your "hobbyist user" arguments ;p

blowdart:
OK so.

Look for a "sub" claim first. If it's present use that, the value and the issuer.
If it's not present look for a ClaimTypes.NameIdentifer property. If it's present use that, the value and the issuer
If that's not present look for ClaimTypes.Upn. If it's present, use that, the value and the issuer

And then if nothing of the above is present, sha256 the entire identity.

Throwing seems rather drastic. Keep in mind we're not using the identity alone, it is combined with a random value.

That work @leastprivilege ?

brockallen:
The one last comment about this is that using the authenticated user's identifier tends to break antiforgery once the user toggles between anon and authenticated (as it did in prior versions of MVC).

blowdart:
Which I'm good with :)

blowdart:
Added bonus - make sure you check all identities on the principal, and have configuration so a developer can specify a claim to use if needed.

[...]

PR #52, opened 2016-02-06
Updated UID generation in DefaultClaimUidExtractor

ajaybhargavb:
Issue - #30 

@blowdart @rynowak

### DIFF
--- a/src/Microsoft.AspNetCore.Antiforgery/Internal/DefaultAntiforgeryTokenGenerator.cs
+++ b/src/Microsoft.AspNetCore.Antiforgery/Internal/DefaultAntiforgeryTokenGenerator.cs
@@ -2,6 +2,7 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
+using System.Collections.Generic;
 using System.Security.Claims;
 using System.Security.Principal;
 using Microsoft.AspNetCore.Http;
@@ -60,16 +61,17 @@ namespace Microsoft.AspNetCore.Antiforgery.Internal
             };
 
             var isIdentityAuthenticated = false;
-            var identity = httpContext.User?.Identity as ClaimsIdentity;
 
             // populate Username and ClaimUid
-            if (identity != null && identity.IsAuthenticated)
+            var authenticatedIdentity = GetAuthenticatedIdentity(httpContext.User);
+            if (authenticatedIdentity != null)
             {
                 isIdentityAuthenticated = true;
-                requestToken.ClaimUid = GetClaimUidBlob(_claimUidExtractor.ExtractClaimUid(identity));
+                requestToken.ClaimUid = GetClaimUidBlob(_claimUidExtractor.ExtractClaimUid(httpContext.User));
+
                 if (requestToken.ClaimUid == null)
                 {
-                    requestToken.Username = identity.Name;
+                    requestToken.Username = authenticatedIdentity.Name;
                 }
             }
 
@@ -87,7 +89,7 @@ namespace Microsoft.AspNetCore.Antiforgery.Internal
                 // Application says user is authenticated, but we have no identifier for the user.
                 throw new InvalidOperationException(
                     Resources.FormatAntiforgeryTokenValidator_AuthenticatedUserWithoutUsername(
-                        identity.GetType(),
+                        authenticatedIdentity.GetType(),
                         nameof(IIdentity.IsAuthenticated),
                         "true",
                         nameof(IIdentity.Name),
@@ -148,13 +150,13 @@ namespace Microsoft.AspNetCore.Antiforgery.Internal
             var currentUsername = string.Empty;
             BinaryBlob currentClaimUid = null;
 
-            var identity = httpContext.User?.Identity as ClaimsIdentity;
-            if (identity != null && identity.IsAuthenticated)
+            var authenticatedIdentity = GetAuthenticatedIdentity(httpContext.User);
+            if (authenticatedIdentity != null)
             {
-                currentClaimUid = GetClaimUidBlob(_claimUidExtractor.ExtractClaimUid(identity));
+                currentClaimUid = GetClaimUidBlob(_claimUidExtractor.ExtractClaimUid(httpContext.User));
                 if (currentClaimUid == null)
                 {
-                    currentUsername = identity.Name ?? string.Empty;
+                    currentUsername = authenticatedIdentity.Name ?? string.Empty;
                 }
             }
 
@@ -199,6 +201,38 @@ namespace Microsoft.AspNetCore.Antiforgery.Internal
             }
 
             return new BinaryBlob(256, Convert.FromBase64String(base64ClaimUid));
+        }
+
+        private static ClaimsIdentity GetAuthenticatedIdentity(ClaimsPrincipal claimsPrincipal)
+        {
+            if (claimsPrincipal == null)
+            {
+                return null;
+            }
+
+            var identitiesList = claimsPrincipal.Identities as List<ClaimsIdentity>;
+            if (identitiesList != null)
+            {
+                for (var i = 0; i < identitiesList.Count; i++)
+                {
+                    if (identitiesList[i].IsAuthenticated)
+                    {
+                        return identitiesList[i];
+                    }
+                }
+            }
+            else
+            {
+                foreach (var identity in claimsPrincipal.Identities)
+                {
+                    if (identity.IsAuthenticated)
+                    {
+                        return identity;
+                    }
+                }
+            }
+
+            return null;
         }
     }
 }
--- a/src/Microsoft.AspNetCore.Antiforgery/Internal/DefaultClaimUidExtractor.cs
+++ b/src/Microsoft.AspNetCore.Antiforgery/Internal/DefaultClaimUidExtractor.cs
@@ -3,7 +3,7 @@
 
 using System;
 using System.Collections.Generic;
-using System.Linq;
+using System.Diagnostics;
 using System.Security.Claims;
 using Microsoft.Extensions.ObjectPool;
 
@@ -22,42 +22,99 @@ namespace Microsoft.AspNetCore.Antiforgery.Internal
         }
 
         /// <inheritdoc />
-        public string ExtractClaimUid(ClaimsIdentity claimsIdentity)
+        public string ExtractClaimUid(ClaimsPrincipal claimsPrincipal)
         {
-            if (claimsIdentity == null || !claimsIdentity.IsAuthenticated)
+            Debug.Assert(claimsPrincipal != null);
+
+            var uniqueIdentifierParameters = GetUniqueIdentifierParameters(claimsPrincipal.Identities);
+            if (uniqueIdentifierParameters == null)
             {
-                // Skip anonymous users
+                // No authenticated identities containing claims found.
                 return null;
             }
 
-            var uniqueIdentifierParameters = GetUniqueIdentifierParameters(claimsIdentity);
             var claimUidBytes = ComputeSha256(uniqueIdentifierParameters);
             return Convert.ToBase64String(claimUidBytes);
         }
 
-        // Internal for testing
-        internal static IEnumerable<string> GetUniqueIdentifierParameters(ClaimsIdentity claimsIdentity)
+        public static IList<string> GetUniqueIdentifierParameters(IEnumerable<ClaimsIdentity> claimsIdentities)
         {
-            var nameIdentifierClaim = claimsIdentity.FindFirst(
-                claim => string.Equals(ClaimTypes.NameIdentifier, claim.Type, StringComparison.Ordinal));
-            if (nameIdentifierClaim != null && !string.IsNullOrEmpty(nameIdentifierClaim.Value))
+            var identitiesList = claimsIdentities as List<ClaimsIdentity>;
+            if (identitiesList == null)
             {
-                return new string[]
-                {
-                    ClaimTypes.NameIdentifier,
-                    nameIdentifierClaim.Value
-                };
+                identitiesList = new List<ClaimsIdentity>(claimsIdentities);
             }
 
-            // We do not understand this ClaimsIdentity, fallback on serializing the entire claims Identity.
-            var claims = claimsIdentity.Claims.ToList();
-            claims.Sort((a, b) => string.Compare(a.Type, b.Type, StringComparison.Ordinal));
-
-            var identifierParameters = new List<string>(claims.Count * 2);
-            foreach (var claim in claims)
+            for (var i = 0; i < identitiesList.Count; i++)
             {
+                var identity = identitiesList[i];
+                if (!identity.IsAuthenticated)
+                {
+                    continue;
+                }
+
+                var subClaim = identity.FindFirst(
+                    claim => string.Equals("sub", claim.Type, StringComparison.Ordinal));
+                if (subClaim != null && !string.IsNullOrEmpty(subClaim.Value))
+                {
+                    return new string[]
+                    {
+                        subClaim.Type,
+                        subClaim.Value,
+                        subClaim.Issuer
+                    };
+                }
+
+                var nameIdentifierClaim = identity.FindFirst(
+                    claim => string.Equals(ClaimTypes.NameIdentifier, claim.Type, StringComparison.Ordinal));
+                if (nameIdentifierClaim != null && !string.IsNullOrEmpty(nameIdentifierClaim.Value))
+                {
+                    return new string[]
+                    {
+                        nameIdentifierClaim.Type,
+                        nameIdentifierClaim.Value,
+                        nameIdentifierClaim.Issuer
+                    };
+                }
+
+                var upnClaim = identity.FindFirst(
+                    claim => string.Equals(ClaimTypes.Upn, claim.Type, StringComparison.Ordinal));
+                if (upnClaim != null && !string.IsNullOrEmpty(upnClaim.Value))
+                {
+                    return new string[]
+                    {
+                        upnClaim.Type,
+                        upnClaim.Value,
+                        upnClaim.Issuer
+                    };
+                }
+            }
+
+            // We do not understand any of the ClaimsIdentity instances, fallback on serializing all claims in every claims Identity.
+            var allClaims = new List<Claim>();
+            for (var i = 0; i < identitiesList.Count; i++)
+            {
+                if (identitiesList[i].IsAuthenticated)
+                {
+                    allClaims.AddRange(identitiesList[i].Claims);
+                }
+            }
+
+            if (allClaims.Count == 0)
+            {
+                // No authenticated identities containing claims found.
+                return null;
+            }
+
+            allClaims.Sort((a, b) => string.Compare(a.Type, b.Type, StringComparison.Ordinal));
+
+            var identifierParameters = new List<string>(allClaims.Count * 3);
+            for (var i = 0; i < allClaims.Count; i++)
+            {
+                var claim = allClaims[i];
                 identifierParameters.Add(claim.Type);
                 identifierParameters.Add(claim.Value);
+                identifierParameters.Add(claim.Issuer);
             }
 
             return identifierParameters;
--- a/src/Microsoft.AspNetCore.Antiforgery/Internal/IClaimUidExtractor.cs
+++ b/src/Microsoft.AspNetCore.Antiforgery/Internal/IClaimUidExtractor.cs
@@ -1,20 +1,21 @@
 // Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+using System.Collections.Generic;
 using System.Security.Claims;
 
 namespace Microsoft.AspNetCore.Antiforgery.Internal
 {
     /// <summary>
-    /// This interface can extract unique identifers for a claims-based identity.
+    /// This interface can extract unique identifers for a <see cref="ClaimsPrincipal"/>.
     /// </summary>
     public interface IClaimUidExtractor
     {
         /// <summary>
         /// Extracts claims identifier.
         /// </summary>
-        /// <param name="identity">The <see cref="ClaimsIdentity"/>.</param>
+        /// <param name="claimsPrincipal">The <see cref="ClaimsPrincipal"/>.</param>
         /// <returns>The claims identifier.</returns>
-        string ExtractClaimUid(ClaimsIdentity identity);
+        string ExtractClaimUid(ClaimsPrincipal claimsPrincipal);
     }
 }
--- a/test/Microsoft.AspNetCore.Antiforgery.Test/Internal/DefaultAntiforgeryTokenGeneratorTest.cs
+++ b/test/Microsoft.AspNetCore.Antiforgery.Test/Internal/DefaultAntiforgeryTokenGeneratorTest.cs
@@ -2,6 +2,7 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
+using System.Collections.Generic;
 using System.Security.Claims;
 using System.Security.Cryptography;
 using Microsoft.AspNetCore.Http.Internal;
@@ -157,7 +158,7 @@ namespace Microsoft.AspNetCore.Antiforgery.Internal
             var expectedClaimUid = new BinaryBlob(256, data);
 
             var mockClaimUidExtractor = new Mock<IClaimUidExtractor>();
-            mockClaimUidExtractor.Setup(o => o.ExtractClaimUid(identity))
+            mockClaimUidExtractor.Setup(o => o.ExtractClaimUid(It.Is<ClaimsPrincipal>(c => c.Identity == identity)))
                                  .Returns(base64ClaimUId);
 
             var tokenProvider = new DefaultAntiforgeryTokenGenerator(
@@ -410,7 +411,7 @@ namespace Microsoft.AspNetCore.Antiforgery.Internal
             };
 
             var mockClaimUidExtractor = new Mock<IClaimUidExtractor>();
-            mockClaimUidExtractor.Setup(o => o.ExtractClaimUid(identity))
+            mockClaimUidExtractor.Setup(o => o.ExtractClaimUid(It.Is<ClaimsPrincipal>(c => c.Identity == identity)))
                                  .Returns((string)null);
 
             var tokenProvider = new DefaultAntiforgeryTokenGenerator(
@@ -448,7 +449,7 @@ namespace Microsoft.AspNetCore.Antiforgery.Internal
 
             var differentToken = new BinaryBlob(256);
             var mockClaimUidExtractor = new Mock<IClaimUidExtractor>();
-            mockClaimUidExtractor.Setup(o => o.ExtractClaimUid(identity))
+            mockClaimUidExtractor.Setup(o => o.ExtractClaimUid(It.Is<ClaimsPrincipal>(c => c.Identity == identity)))
                                  .Returns(Convert.ToBase64String(differentToken.GetData()));
 
             var tokenProvider = new DefaultAntiforgeryTokenGenerator(
@@ -590,7 +591,7 @@ namespace Microsoft.AspNetCore.Antiforgery.Internal
             };
 
             var mockClaimUidExtractor = new Mock<IClaimUidExtractor>();
-            mockClaimUidExtractor.Setup(o => o.ExtractClaimUid(identity))
+            mockClaimUidExtractor.Setup(o => o.ExtractClaimUid(It.Is<ClaimsPrincipal>(c => c.Identity == identity)))
                                  .Returns(Convert.ToBase64String(fieldtoken.ClaimUid.GetData()));
 
             var tokenProvider = new DefaultAntiforgeryTokenGenerator(
--- a/test/Microsoft.AspNetCore.Antiforgery.Test/Internal/DefaultClaimUidExtractorTest.cs
+++ b/test/Microsoft.AspNetCore.Antiforgery.Test/Internal/DefaultClaimUidExtractorTest.cs
@@ -2,6 +2,8 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
+using System.Collections.Generic;
+using System.Diagnostics;
 using System.Linq;
 using System.Security.Claims;
 using Microsoft.Extensions.ObjectPool;
@@ -16,19 +18,6 @@ namespace Microsoft.AspNetCore.Antiforgery.Internal
             new DefaultObjectPoolProvider().Create(new AntiforgerySerializationContextPooledObjectPolicy());
 
         [Fact]
-        public void ExtractClaimUid_NullIdentity()
-        {
-            // Arrange
-            var extractor = new DefaultClaimUidExtractor(_pool);
-
-            // Act
-            var claimUid = extractor.ExtractClaimUid(null);
-
-            // Assert
-            Assert.Null(claimUid);
-        }
-
-        [Fact]
         public void ExtractClaimUid_Unauthenticated()
         {
             // Arrange
@@ -39,7 +28,7 @@ namespace Microsoft.AspNetCore.Antiforgery.Internal
                         .Returns(false);
 
             // Act
-            var claimUid = extractor.ExtractClaimUid(mockIdentity.Object);
+            var claimUid = extractor.ExtractClaimUid(new ClaimsPrincipal(mockIdentity.Object));
 
             // Assert
             Assert.Null(claimUid);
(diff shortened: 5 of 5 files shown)


src/Microsoft.AspNetCore.Antiforgery/Internal/DefaultClaimUidExtractor.cs
```diff
@@ -38,6 +38,17 @@ public string ExtractClaimUid(ClaimsIdentity claimsIdentity)
...
         internal static IEnumerable<string> GetUniqueIdentifierParameters(ClaimsIdentity claimsIdentity)
         {
+            var subClaim = claimsIdentity.FindFirst(
+                claim => string.Equals("sub", claim.Type, StringComparison.Ordinal));
+            if (subClaim != null && !string.IsNullOrEmpty(subClaim.Value))
+            {
+                return new string[]
+                {
+                    "sub",
+                    subClaim.Value
```
rynowak: This comment https://github.com/aspnet/Antiforgery/issues/30#issuecomment-180546654 mentions the claim and issuer... @blowdart want to explain exactly what to do?

src/Microsoft.AspNetCore.Antiforgery/Internal/DefaultClaimUidExtractor.cs
```diff
@@ -38,6 +38,17 @@ public string ExtractClaimUid(ClaimsIdentity claimsIdentity)
         // Internal for testing
         internal static IEnumerable<string> GetUniqueIdentifierParameters(ClaimsIdentity claimsIdentity)
```
rynowak: @blowdart - does it matter that we're operating on a single `ClaimsIdentity`?

rynowak:
:watch:

src/Microsoft.AspNetCore.Antiforgery/Internal/DefaultClaimUidExtractor.cs
blowdart: You'd take the claim value, and the claim issuer (the Issuer property on the claim), and safely concatenate them (basically you do len(value)+value+len(issuer)+issuer), then either put that in the token, or a SHA256 hash of the concatenated value.

This basically means that if, for example, facebook and google issued the same sub claim value, you could still tell them apart, because the issuer is different.

src/Microsoft.AspNetCore.Antiforgery/Internal/DefaultClaimUidExtractor.cs
blowdart: We should only be looking at identities where the IsAuthenticated flag is true. If there are multiple ones, then either we choose the first which contains one of the claims we want, or perhaps look through all of them. I think using the first would be better, as we create "half" identities we throw away during social authentication.

test/Microsoft.AspNetCore.Antiforgery.Test/Internal/DefaultClaimUidExtractorTest.cs
```diff
@@ -111,5 +111,63 @@ public void DefaultUniqueClaimTypes_Present()
...
+                "nameIdentifierValue"
+            }, uniqueIdentifierParameters);
+        }
+
+        [Fact]
+        public void GetUniqueIdentifierParameters_UsesUpnIfPresent()
+        {
+            // Arrange
+            var identity = new ClaimsIdentity();
+            identity.AddClaim(new Claim("fooClaim", "fooClaimValue"));
```
dougbu: Nit: Please avoid "foo" and similar

src/Microsoft.AspNetCore.Antiforgery/Internal/DefaultClaimUidExtractor.cs
dougbu: @blowdart the code currently never goes beyond the first `ClaimsIdentity`. Caller uses `User?.Identity as ClaimsIdentity` and not a search through `User?.Identities`. Is that significant?

dougbu: BTW `IsAuthenticated` is checked both just above (in `ExtractClaimUid()`) and in the `DefaultAntiforgeryTokenGenerator` calling code.

src/Microsoft.AspNetCore.Antiforgery/Internal/DefaultClaimUidExtractor.cs
dougbu: Well these 2-`string` arrays are run through a SHA256 hash soon after this method returns. Questions:
- Should they be 3-`string` arrays, adding the issuer?
- Use `Issuer` or `OriginalIssuer` from the `Claim`?

test/Microsoft.AspNetCore.Antiforgery.Test/Internal/DefaultClaimUidExtractorTest.cs
Eilon: Our tests use `foo` in thousands of places, what's wrong with it? It at least makes it clear that it's an arbitrary value...

ajaybhargavb:
Updated.

ajaybhargavb:
Need to add more tests. Will do it once the updated design is accepted.

src/Microsoft.AspNetCore.Antiforgery/Internal/DefaultAntiforgeryTokenGenerator.cs
```diff
@@ -66,7 +66,12 @@ public AntiforgeryToken GenerateCookieToken()
             if (identity != null && identity.IsAuthenticated)
```
rynowak: Should we still check this now that you're checking `User?.Identities`? Seems rendundant

src/Microsoft.AspNetCore.Antiforgery/Internal/DefaultClaimUidExtractor.cs
```diff
@@ -22,42 +22,89 @@ public DefaultClaimUidExtractor(ObjectPool<AntiforgerySerializationContext> pool
...
+            var allClaims = new List<Claim>();
+            foreach (var identity in claimsIdentities)
+            {
+                if (!identity.IsAuthenticated)
+                {
+                    continue;
+                }
+
+                var subClaim = identity.FindFirst(
+                claim => string.Equals("sub", claim.Type, StringComparison.Ordinal));
```
rynowak: fix indent

src/Microsoft.AspNetCore.Antiforgery/Internal/DefaultClaimUidExtractor.cs
```diff
@@ -22,42 +22,89 @@ public DefaultClaimUidExtractor(ObjectPool<AntiforgerySerializationContext> pool
...
             }
 
-            var uniqueIdentifierParameters = GetUniqueIdentifierParameters(claimsIdentity);
             var claimUidBytes = ComputeSha256(uniqueIdentifierParameters);
             return Convert.ToBase64String(claimUidBytes);
         }
 
         // Internal for testing
-        internal static IEnumerable<string> GetUniqueIdentifierParameters(ClaimsIdentity claimsIdentity)
+        internal static IEnumerable<string> GetUniqueIdentifierParameters(IEnumerable<ClaimsIdentity> claimsIdentities)
```
rynowak: Return this as IList<>

src/Microsoft.AspNetCore.Antiforgery/Internal/DefaultClaimUidExtractor.cs
```diff
@@ -22,42 +22,89 @@ public DefaultClaimUidExtractor(ObjectPool<AntiforgerySerializationContext> pool
...
-            var nameIdentifierClaim = claimsIdentity.FindFirst(
-                claim => string.Equals(ClaimTypes.NameIdentifier, claim.Type, StringComparison.Ordinal));
-            if (nameIdentifierClaim != null && !string.IsNullOrEmpty(nameIdentifierClaim.Value))
+            if (claimsIdentities == null)
             {
-                return new string[]
+                return null;
+            }
+
+            var allClaims = new List<Claim>();
```
rynowak: Don't allocate the list unless needed, since that's the uncommon case

src/Microsoft.AspNetCore.Antiforgery/Internal/DefaultClaimUidExtractor.cs
```diff
@@ -22,42 +22,89 @@ public DefaultClaimUidExtractor(ObjectPool<AntiforgerySerializationContext> pool
...
+                {
+                    return new string[]
+                    {
+                        upnClaim.Type,
+                        upnClaim.Value,
+                        upnClaim.Issuer
+                    };
+                }
+
+                allClaims.AddRange(identity.Claims);
```
rynowak: I'd even put this in a separate block and do a second pass through the identities if needed.

src/Microsoft.AspNetCore.Antiforgery/Internal/IClaimUidExtractor.cs
```diff
@@ -1,20 +1,21 @@
...
     public interface IClaimUidExtractor
     {
         /// <summary>
         /// Extracts claims identifier.
         /// </summary>
-        /// <param name="identity">The <see cref="ClaimsIdentity"/>.</param>
+        /// <param name="identities">The list of <see cref="ClaimsIdentity"/>.</param>
         /// <returns>The claims identifier.</returns>
-        string ExtractClaimUid(ClaimsIdentity identity);
+        string ExtractClaimUid(IEnumerable<ClaimsIdentity> identities);
```
rynowak: Why not just have this take a `ClaimsPrincipal`? It's a bit unclear what the semantics are supposed to be for "list of identities".

src/Microsoft.AspNetCore.Antiforgery/Internal/DefaultClaimUidExtractor.cs
rynowak: Just make this public. There's no harm in it, this is an internal type.

test/Microsoft.AspNetCore.Antiforgery.Test/Internal/DefaultClaimUidExtractorTest.cs
```diff
@@ -97,18 +99,80 @@ public void DefaultUniqueClaimTypes_NotPresent_SerializesAllClaimTypes()
...
             Assert.Equal(new string[]
             {
                 ClaimTypes.NameIdentifier,
                 "nameIdentifierValue",
+                "LOCAL AUTHORITY",
+            }, uniqueIdentifierParameters);
+        }
+
+        [Fact]
+        public void GetUniqueIdentifierParameters_PrefersSubClaimOverNameIdentifierAndUpn()
```
rynowak: Need to see more tests with unauthenticated identities, and with multiple authenticated identities for the ultimate fallback case.

rynowak:
:watch:

src/Microsoft.AspNetCore.Antiforgery/Internal/DefaultAntiforgeryTokenGenerator.cs
dougbu: Also, why does caller pass in `httpContext.User.Identity` when this code turns around and uses `httpContext.User?.Identities`? Another reason to start w/ the `ClaimsPrinciple` as @rynowak suggested [below](https://github.com/aspnet/Antiforgery/pull/52/files#r52238020).

dougbu: Note that if `Identity` is equal to the first member of `Identities` where `IsAuthenticated` is `true`, existing code was :ok:. (I'm not sure whether "primary" in the `Identity` doc comment means `IsAuthenticated` or something else.)

src/Microsoft.AspNetCore.Antiforgery/Internal/DefaultClaimUidExtractor.cs
```diff
@@ -22,42 +22,89 @@ public DefaultClaimUidExtractor(ObjectPool<AntiforgerySerializationContext> pool
...
         }
 
         // Internal for testing
-        internal static IEnumerable<string> GetUniqueIdentifierParameters(ClaimsIdentity claimsIdentity)
+        internal static IEnumerable<string> GetUniqueIdentifierParameters(IEnumerable<ClaimsIdentity> claimsIdentities)
         {
-            var nameIdentifierClaim = claimsIdentity.FindFirst(
-                claim => string.Equals(ClaimTypes.NameIdentifier, claim.Type, StringComparison.Ordinal));
-            if (nameIdentifierClaim != null && !string.IsNullOrEmpty(nameIdentifierClaim.Value))
+            if (claimsIdentities == null)
```
dougbu: Can the `IEnumerable<T>` be `null` in real scenarios?

test/Microsoft.AspNetCore.Antiforgery.Test/Internal/DefaultAntiforgeryTokenGeneratorTest.cs
```diff
@@ -157,7 +158,7 @@ public void GenerateRequestToken_ClaimsBasedIdentity()
             var expectedClaimUid = new BinaryBlob(256, data);
 
             var mockClaimUidExtractor = new Mock<IClaimUidExtractor>();
-            mockClaimUidExtractor.Setup(o => o.ExtractClaimUid(identity))
+            mockClaimUidExtractor.Setup(o => o.ExtractClaimUid(It.IsAny<IEnumerable<ClaimsIdentity>>()))
```
dougbu: Losing a few failure cases here. Test would fail when the method was not passed the exact `ClaimsIdentity` expected.

dougbu:
:watch:

src/Microsoft.AspNetCore.Antiforgery/Internal/DefaultAntiforgeryTokenGenerator.cs
ajaybhargavb: I looked at the code. It does not check for `IsAuthenticated`. It simply selects first non-null identity. https://github.com/dotnet/corefx/blob/master/src/System.Security.Claims/src/System/Security/Claims/ClaimsPrincipal.cs

src/Microsoft.AspNetCore.Antiforgery/Internal/DefaultClaimUidExtractor.cs
ajaybhargavb: Not sure. But I guess returning null is better than just letting it blow up.

ajaybhargavb:
Updated.

src/Microsoft.AspNetCore.Antiforgery/Internal/DefaultAntiforgeryTokenGenerator.cs
```diff
@@ -60,16 +61,17 @@ public AntiforgeryToken GenerateCookieToken()
             };
 
             var isIdentityAuthenticated = false;
-            var identity = httpContext.User?.Identity as ClaimsIdentity;
+            var authenticatedIdentity = httpContext.User?.Identities?.FirstOrDefault(c => c.IsAuthenticated == true);
```
dougbu: We're trying hard to _optimize_ Antiforgery. Don't introduce Linq here.

src/Microsoft.AspNetCore.Antiforgery/Internal/DefaultClaimUidExtractor.cs
```diff
@@ -22,42 +22,94 @@ public DefaultClaimUidExtractor(ObjectPool<AntiforgerySerializationContext> pool
...
+                    {
+                        upnClaim.Type,
+                        upnClaim.Value,
+                        upnClaim.Issuer
+                    };
+                }
+            }
+
+            // We do not understand any of the ClaimsIdentity instances, fallback on serializing all claims in every claims Identity.
+            var allClaims = new List<Claim>();
```
dougbu: Get rid of this extra `List<T>` allocation. Just use `identifierParameters` but add the `IsAuthenticated` check inside the loop. Suspect most identities will be authenticated but should probably check w/ @blowdart about expected collection counts versus expected counts of authenticated entries.

src/Microsoft.AspNetCore.Antiforgery/Internal/DefaultClaimUidExtractor.cs
```diff
@@ -22,42 +22,94 @@ public DefaultClaimUidExtractor(ObjectPool<AntiforgerySerializationContext> pool
...
+            {
+                if (identity.IsAuthenticated)
+                {
+                    allClaims.AddRange(identity.Claims);
+                }
+            }
+
+            if (allClaims.Count == 0)
+            {
+                // No authenticated identities found.
```
dougbu: How is this possible? The callers make sure at least one identity is authenticated.

src/Microsoft.AspNetCore.Antiforgery/Internal/DefaultClaimUidExtractor.cs
blowdart: It's an edge case to get an unauthenticated identity, we use it for social logins, but it does happen. Usually, like 99% of the time we'll have a single authenticated identity.

src/Microsoft.AspNetCore.Antiforgery/Internal/DefaultClaimUidExtractor.cs
dougbu: We don't add these kind of checks just because. It may mislead developers down the road if it's not a real scenario. Check.

dougbu: Actually, this isn't a real scenario since the caller checks exactly this condition. Just `Debug.Assert()`.

src/Microsoft.AspNetCore.Antiforgery/Internal/DefaultClaimUidExtractor.cs
ajaybhargavb: Yes. But we were checking of `IsAuthenticated` multiple times even before this change. Maybe for cases when someone uses a custom `IAntiforgeryTokenGenerator`.

src/Microsoft.AspNetCore.Antiforgery/Internal/DefaultAntiforgeryTokenGenerator.cs
dougbu: Also,
- Can `Identities` really be `null`? Check the real code and don't check here unless it's a real scenario.
- Lots of boilerplate around `authenticatedIdentity` repeated in two methods. Suggest a helper method.

dougbu:
:watch:

src/Microsoft.AspNetCore.Antiforgery/Internal/DefaultAntiforgeryTokenGenerator.cs
ajaybhargavb: > Can Identities really be null?
You are right. I can never be null. An empty list is always allocated.

src/Microsoft.AspNetCore.Antiforgery/Internal/DefaultClaimUidExtractor.cs
ajaybhargavb: > Get rid of this extra List<T> allocation.

Spoke offline. Cannot do this because we sort the claims before adding it to the `identityParameters` list.

src/Microsoft.AspNetCore.Antiforgery/Internal/DefaultClaimUidExtractor.cs
ajaybhargavb: Actually this is possible because we check for an authenticated identity but that identity can contain 0 claims.

ajaybhargavb:
Updated.

test/Microsoft.AspNetCore.Antiforgery.Test/Internal/DefaultClaimUidExtractorTest.cs
ajaybhargavb: Added.

src/Microsoft.AspNetCore.Antiforgery/Internal/DefaultAntiforgeryTokenGenerator.cs
```diff
@@ -60,16 +61,17 @@ public AntiforgeryToken GenerateCookieToken()
             };
 
             var isIdentityAuthenticated = false;
-            var identity = httpContext.User?.Identity as ClaimsIdentity;
+            var authenticatedIdentity = httpContext.User != null ? GetAuthenticatedIdentity(httpContext.User) : null;
```
dougbu: Remove the ternary and have GetAuthenticatedIdentity() return `null` when passed `null`.

src/Microsoft.AspNetCore.Antiforgery/Internal/DefaultClaimUidExtractor.cs
```diff
@@ -22,42 +23,94 @@ public DefaultClaimUidExtractor(ObjectPool<AntiforgerySerializationContext> pool
...
 
         /// <inheritdoc />
-        public string ExtractClaimUid(ClaimsIdentity claimsIdentity)
+        public string ExtractClaimUid(ClaimsPrincipal claimsPrincipal)
         {
-            if (claimsIdentity == null || !claimsIdentity.IsAuthenticated)
+            if (claimsPrincipal == null)
+            {
+                throw new ArgumentNullException(nameof(claimsPrincipal));
+            }
```
dougbu: This is not a public API. Just `Debug.Assert()` when you want to check a parameter.

src/Microsoft.AspNetCore.Antiforgery/Internal/DefaultClaimUidExtractor.cs
```diff
@@ -22,42 +23,94 @@ public DefaultClaimUidExtractor(ObjectPool<AntiforgerySerializationContext> pool
...
         }
 
-        // Internal for testing
-        internal static IEnumerable<string> GetUniqueIdentifierParameters(ClaimsIdentity claimsIdentity)
+        public static IList<string> GetUniqueIdentifierParameters(IEnumerable<ClaimsIdentity> claimsIdentities)
         {
-            var nameIdentifierClaim = claimsIdentity.FindFirst(
-                claim => string.Equals(ClaimTypes.NameIdentifier, claim.Type, StringComparison.Ordinal));
-            if (nameIdentifierClaim != null && !string.IsNullOrEmpty(nameIdentifierClaim.Value))
+            foreach (var identity in claimsIdentities)
```
dougbu: Not sure but I suspect going through the `claimsIdentities` twice makes it worthwhile saving the enumerable to an array. Then again is this another place where the enumerable is always backed by a `List<T>`.

src/Microsoft.AspNetCore.Antiforgery/Internal/DefaultClaimUidExtractor.cs
dougbu: Agree though this will often perform worse than previous code because `ToList()` creates a new `List<T>` of the correct size -- no need for expansion.

src/Microsoft.AspNetCore.Antiforgery/Internal/DefaultClaimUidExtractor.cs
```diff
@@ -22,42 +23,94 @@ public DefaultClaimUidExtractor(ObjectPool<AntiforgerySerializationContext> pool
...
+            var allClaims = new List<Claim>();
+            foreach (var identity in claimsIdentities)
             {
-                return new string[]
+                if (identity.IsAuthenticated)
                 {
-                    ClaimTypes.NameIdentifier,
-                    nameIdentifierClaim.Value
-                };
+                    allClaims.AddRange(identity.Claims);
```
dougbu: The new code adds all claims for all identities in the given `Enumerable<ClaimsIdentity>`. The old code only added claims for one identity. Should `break` after this line if `allClaims.Count != 0` to preserve that behaviour.

test/Microsoft.AspNetCore.Antiforgery.Test/Internal/DefaultClaimUidExtractorTest.cs
```diff
@@ -97,18 +97,133 @@ public void DefaultUniqueClaimTypes_NotPresent_SerializesAllClaimTypes()
...
+            {
+                ClaimTypes.Country,
+                "countryValue",
+                "LOCAL AUTHORITY",
+                ClaimTypes.Email,
+                "email@domain.com",
+                "LOCAL AUTHORITY",
+                ClaimTypes.Name,
+                "claimName",
+                "LOCAL AUTHORITY",
```
dougbu: This test demonstrates the behaviour I complained about [here](https://github.com/aspnet/Antiforgery/pull/52/files#r52409185). Should never mix claims from multiple identities in the same UID.

src/Microsoft.AspNetCore.Antiforgery/Internal/DefaultClaimUidExtractor.cs
dougbu: My suggestion is to stick with `ToList()` then do a `for` through the collection and `.Remove()` the (rare) unauthenticated identity.

dougbu:
:watch:

src/Microsoft.AspNetCore.Antiforgery/Internal/DefaultClaimUidExtractor.cs
ajaybhargavb: @blowdart wanted this behavior if I am not wrong. We need to squeeze the claims of all identities together if we can't find the claim type we want.

blowdart: Yup. This makes me a happy camper.

test/Microsoft.AspNetCore.Antiforgery.Test/Internal/DefaultClaimUidExtractorTest.cs
ajaybhargavb: https://github.com/aspnet/Antiforgery/pull/52#discussion_r52498650

src/Microsoft.AspNetCore.Antiforgery/Internal/DefaultClaimUidExtractor.cs
ajaybhargavb: Since we are adding claims from multiple identities https://github.com/aspnet/Antiforgery/pull/52#discussion_r52498650, this can't be done with a simple `ToList()`. I don't think this will be an improvement.

rynowak: Let's not go crazy please without evidence that there is a problem. Let's fix the incorrectness of this code and then figure out if it even needs to be optimized.

ajaybhargavb:
Updated.

src/Microsoft.AspNetCore.Antiforgery/Internal/IClaimUidExtractor.cs
```diff
@@ -1,20 +1,21 @@
...
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+using System.Collections.Generic;
 using System.Security.Claims;
 
 namespace Microsoft.AspNetCore.Antiforgery.Internal
 {
     /// <summary>
-    /// This interface can extract unique identifers for a claims-based identity.
+    /// This interface can extract unique identifers for a list of claims-based identities.
```
rynowak: For a `ClaimsPrincipal`

rynowak:
:shipit: from me. Get a signoff of @dougbu and @blowdart also

src/Microsoft.AspNetCore.Antiforgery/Internal/DefaultClaimUidExtractor.cs
```diff
@@ -22,42 +23,99 @@ public DefaultClaimUidExtractor(ObjectPool<AntiforgerySerializationContext> pool
...
-        public string ExtractClaimUid(ClaimsIdentity claimsIdentity)
+        public string ExtractClaimUid(ClaimsPrincipal claimsPrincipal)
         {
-            if (claimsIdentity == null || !claimsIdentity.IsAuthenticated)
+            Debug.Assert(claimsPrincipal != null);
+
+            var uniqueIdentifierParameters = GetUniqueIdentifierParameters(claimsPrincipal.Identities);
+            if (uniqueIdentifierParameters == null)
             {
                 // Skip anonymous users
```
dougbu: Fix this comment. Never reach here if the principal isn't logged in somehow.

src/Microsoft.AspNetCore.Antiforgery/Internal/DefaultClaimUidExtractor.cs
```diff
@@ -22,42 +23,99 @@ public DefaultClaimUidExtractor(ObjectPool<AntiforgerySerializationContext> pool
...
-        internal static IEnumerable<string> GetUniqueIdentifierParameters(ClaimsIdentity claimsIdentity)
+        public static IList<string> GetUniqueIdentifierParameters(IEnumerable<ClaimsIdentity> claimsIdentities)
         {
-            var nameIdentifierClaim = claimsIdentity.FindFirst(
-                claim => string.Equals(ClaimTypes.NameIdentifier, claim.Type, StringComparison.Ordinal));
-            if (nameIdentifierClaim != null && !string.IsNullOrEmpty(nameIdentifierClaim.Value))
+            var identitiesList = claimsIdentities as List<ClaimsIdentity>;
+            if (identitiesList == null)
+            {
+                identitiesList = claimsIdentities.ToList();
```
dougbu: FYI `enumerable.ToList()` is pretty much identical to `new List<T>(enumerable)`. No need to change unless it allows you to remove the `using` for Linq; then would be good.

test/Microsoft.AspNetCore.Antiforgery.Test/Internal/DefaultClaimUidExtractorTest.cs
```diff
@@ -16,16 +17,13 @@ public class DefaultClaimUidExtractorTest
...
             // Arrange
             var extractor = new DefaultClaimUidExtractor(_pool);
 
-            // Act
-            var claimUid = extractor.ExtractClaimUid(null);
-
-            // Assert
-            Assert.Null(claimUid);
+            // Act & Assert
+            Assert.Throws<ArgumentNullException>(() => extractor.ExtractClaimUid(null));
```
dougbu: Doesn't this hit a `Debug.Assert()` failure? Just remove this test if so, maybe even regardless since it's not a real scenario.

ajaybhargavb: Yes. I removed the test. Just missed the change in this commit.

test/Microsoft.AspNetCore.Antiforgery.Test/Internal/DefaultClaimUidExtractorTest.cs
```diff
@@ -97,18 +97,133 @@ public void DefaultUniqueClaimTypes_NotPresent_SerializesAllClaimTypes()
...
+            Assert.Equal(new string[]
+            {
+                ClaimTypes.Upn,
+                "upnClaimValue",
+                "LOCAL AUTHORITY",
+            }, uniqueIdentifierParameters);
+        }
+
+        [Fact]
+        public void GetUniqueIdentifierParameters_MultipleIdentities_UsesOnlyAuthenticatedIdentities()
```
dougbu: Add test or tests showing that extractor finds the first authenticated `ClaimsIdentity` with a name or UPN claim, not a later one with a sub claim. Same for a `ClaimsIdentity with a UPN claim which is chosen over a later one with a sub (or name though testing both seems redundant) claim.

test/Microsoft.AspNetCore.Antiforgery.Test/Internal/DefaultClaimUidExtractorTest.cs
dougbu: :+1:

dougbu:
:shipit: with a couple more tests. Suggest walking @blowdart though the tests and getting his agreement on the new behaviour.

test/Microsoft.AspNetCore.Antiforgery.Test/Internal/DefaultClaimUidExtractorTest.cs
ajaybhargavb: Added.

ajaybhargavb:
Updated and Rebased.

ajaybhargavb:
Ping @blowdart

blowdart:
Works for me. Might need more kittens though.

PR #53, opened 2016-02-06
Rybrande/add logging

### LINKED ISSUE #44: Add logging to antiforgery
We should add logging

Some ideas:
- Validating a token (verbose)
- Validation failure / missing token (warning)

Thinking warning is right for failures, because generally this is a app configuration issue.

ryanbrandenburg:
Fixes #44

### DIFF
--- a/src/Microsoft.AspNetCore.Antiforgery/Internal/DefaultAntiforgery.cs
+++ b/src/Microsoft.AspNetCore.Antiforgery/Internal/DefaultAntiforgery.cs
@@ -5,6 +5,7 @@ using System;
 using System.Diagnostics;
 using System.Threading.Tasks;
 using Microsoft.AspNetCore.Http;
+using Microsoft.Extensions.Logging;
 using Microsoft.Extensions.Options;
 
 namespace Microsoft.AspNetCore.Antiforgery.Internal
@@ -19,17 +20,20 @@ namespace Microsoft.AspNetCore.Antiforgery.Internal
         private readonly IAntiforgeryTokenGenerator _tokenGenerator;
         private readonly IAntiforgeryTokenSerializer _tokenSerializer;
         private readonly IAntiforgeryTokenStore _tokenStore;
+        private readonly ILogger<DefaultAntiforgery> _logger;
 
         public DefaultAntiforgery(
             IOptions<AntiforgeryOptions> antiforgeryOptionsAccessor,
             IAntiforgeryTokenGenerator tokenGenerator,
             IAntiforgeryTokenSerializer tokenSerializer,
-            IAntiforgeryTokenStore tokenStore)
+            IAntiforgeryTokenStore tokenStore,
+            ILoggerFactory loggerFactory)
         {
             _options = antiforgeryOptionsAccessor.Value;
             _tokenGenerator = tokenGenerator;
             _tokenSerializer = tokenSerializer;
             _tokenStore = tokenStore;
+            _logger = loggerFactory.CreateLogger<DefaultAntiforgery>();
         }
 
         /// <inheritdoc />
@@ -46,6 +50,11 @@ namespace Microsoft.AspNetCore.Antiforgery.Internal
             if (tokenSet.IsNewCookieToken)
             {
                 SaveCookieTokenAndHeader(httpContext, tokenSet.CookieToken);
+                _logger.NewCookieToken();
+            }
+            else
+            {
+                _logger.ReusedCookieToken();
             }
 
             return Serialize(tokenSet);
@@ -76,8 +85,14 @@ namespace Microsoft.AspNetCore.Antiforgery.Internal
             CheckSSLConfig(httpContext);
 
             var tokens = await _tokenStore.GetRequestTokensAsync(httpContext);
-            if (tokens.CookieToken == null || tokens.RequestToken == null)
+            if (tokens.CookieToken == null)
             {
+                _logger.MissingCookieToken(_options.CookieName);
+                return false;
+            }
+            if (tokens.RequestToken == null)
+            {
+                _logger.MissingRequestToken(_options.FormFieldName, _options.HeaderName);
                 return false;
             }
 
@@ -87,11 +102,22 @@ namespace Microsoft.AspNetCore.Antiforgery.Internal
 
             // Validate
             string message;
-            return _tokenGenerator.TryValidateTokenSet(
+            var result = _tokenGenerator.TryValidateTokenSet(
                 httpContext,
                 deserializedCookieToken,
                 deserializedRequestToken,
                 out message);
+
+            if (result)
+            {
+                _logger.ValidatedAntiforgeryToken();
+            }
+            else
+            {
+                _logger.ValidationFailed(message);
+            }
+
+            return result;
         }
 
         /// <inheritdoc />
@@ -133,6 +159,8 @@ namespace Microsoft.AspNetCore.Antiforgery.Internal
             }
 
             ValidateTokens(httpContext, tokens);
+
+            _logger.ValidatedAntiforgeryToken();
         }
 
         private void ValidateTokens(HttpContext httpContext, AntiforgeryTokenSet antiforgeryTokenSet)
--- a/test/Microsoft.AspNetCore.Antiforgery.Test/Internal/DefaultAntiforgeryTest.cs
+++ b/test/Microsoft.AspNetCore.Antiforgery.Test/Internal/DefaultAntiforgeryTest.cs
@@ -6,6 +6,9 @@ using System.Security.Claims;
 using System.Threading.Tasks;
 using Microsoft.AspNetCore.Http;
 using Microsoft.AspNetCore.Http.Internal;
+using Microsoft.AspNetCore.Testing;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Logging;
 using Microsoft.Extensions.Options;
 using Moq;
 using Xunit;
@@ -18,13 +21,12 @@ namespace Microsoft.AspNetCore.Antiforgery.Internal
         public async Task ChecksSSL_ValidateRequestAsync_Throws()
         {
             // Arrange
-            var httpContext = new DefaultHttpContext();
+            var httpContext = GetHttpContext();
             var options = new AntiforgeryOptions()
             {
                 RequireSsl = true
             };
-
-            var antiforgery = GetAntiforgery(options);
+            var antiforgery = GetAntiforgery(httpContext, options);
 
             // Act & Assert
             var exception = await Assert.ThrowsAsync<InvalidOperationException>(
@@ -39,13 +41,13 @@ namespace Microsoft.AspNetCore.Antiforgery.Internal
         public async Task ChecksSSL_IsRequestValidAsync_Throws()
         {
             // Arrange
-            var httpContext = new DefaultHttpContext();
+            var httpContext = GetHttpContext();
             var options = new AntiforgeryOptions()
             {
                 RequireSsl = true
             };
 
-            var antiforgery = GetAntiforgery(options);
+            var antiforgery = GetAntiforgery(httpContext, options);
 
             // Act & Assert
             var exception = await Assert.ThrowsAsync<InvalidOperationException>(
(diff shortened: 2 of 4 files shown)


src/Microsoft.AspNetCore.Antiforgery/Internal/DefaultAntiforgery.cs
```diff
@@ -170,7 +180,10 @@ private void ValidateTokens(HttpContext httpContext, AntiforgeryTokenSet antifor
                 deserializedRequestToken,
                 out message))
             {
-                throw new AntiforgeryValidationException(message);
+                var ex = new AntiforgeryValidationException(message);
```
dougbu: Should probably spell out `exception` and leave `ex` for tests. But I see some `catch (Exception ex)` in other repos so obviously this isn't a hard line.

test/Microsoft.AspNetCore.Antiforgery.Test/Internal/DefaultAntiforgeryTest.cs
```diff
@@ -630,4 +642,4 @@ private class TestOptionsManager : IOptions<AntiforgeryOptions>
             public AntiforgeryOptions Value { get; set; } = new AntiforgeryOptions();
         }
     }
-}
\ No newline at end of file
+}
```
dougbu: Wow, what a horrendous thing to do to this poor file! :trollface:

src/Microsoft.AspNetCore.Antiforgery/Internal/DefaultAntiforgeryLoggerExtensions.cs
```diff
@@ -0,0 +1,77 @@
...
+                4,
+                "The required antiforgery form field '{FormFieldName}' is not present.");
+            _missingHeaderToken = LoggerMessage.Define<string>(
+                LogLevel.Warning,
+                5,
+                "The required antiforgery header value '{HeaderName}' is not present.");
+            _missingRequestToken = LoggerMessage.Define<string, string>(
+                LogLevel.Warning,
+                6,
+                "The required antiforgery request token was not provided in either form field '{FormFieldName}'"
```
dougbu: Space at the end or start next part w/ a space.

src/Microsoft.AspNetCore.Antiforgery/Internal/DefaultAntiforgeryLoggerExtensions.cs
```diff
@@ -0,0 +1,77 @@
...
+        private static readonly Action<ILogger, string, Exception> _missingFormToken;
+        private static readonly Action<ILogger, string, Exception> _missingHeaderToken;
+        private static readonly Action<ILogger, string, string, Exception> _missingRequestToken;
+
+        static DefaultAntiforgeryLoggerExtensions()
+        {
+            _validationFailed = LoggerMessage.Define<string>(
+                LogLevel.Warning,
+                1,
+                "Antiforgery validation failed with this message '{Message}'");
```
dougbu: Remove "this"

src/Microsoft.AspNetCore.Antiforgery/Internal/DefaultAntiforgeryLoggerExtensions.cs
```diff
@@ -0,0 +1,77 @@
...
+                3,
+                "The required antiforgery cookie '{CookieName}' is not present.");
+            _missingFormToken = LoggerMessage.Define<string>(
+                LogLevel.Warning,
+                4,
+                "The required antiforgery form field '{FormFieldName}' is not present.");
+            _missingHeaderToken = LoggerMessage.Define<string>(
+                LogLevel.Warning,
+                5,
+                "The required antiforgery header value '{HeaderName}' is not present.");
```
dougbu: Minor but "value" doesn't add much here or on [line 44](https://github.com/aspnet/Antiforgery/pull/53/files#diff-352f92904f2079f3a50332fd4ffe9a93R44).

src/Microsoft.AspNetCore.Antiforgery/Internal/DefaultAntiforgeryLoggerExtensions.cs
```diff
@@ -0,0 +1,77 @@
...
+            _missingRequestToken = LoggerMessage.Define<string, string>(
+                LogLevel.Warning,
+                6,
+                "The required antiforgery request token was not provided in either form field '{FormFieldName}'"
+                    + "or header value '{HeaderName}'.");
+        }
+
+        public static void ValidationFailed(this ILogger logger, AntiforgeryValidationException exception)
+        {
+            _validationFailed(logger, exception.Message, exception);
```
dougbu: Is this the kind of code we recently decided should use `exception.ToString()`?

src/Microsoft.AspNetCore.Antiforgery/Internal/DefaultAntiforgery.cs
```diff
@@ -170,7 +180,10 @@ private void ValidateTokens(HttpContext httpContext, AntiforgeryTokenSet antifor
...
+
+                _logger.ValidationFailed(ex);
+                throw ex;
             }
```
dougbu: Should this method log if it succeeds? See a log of begin / end pairs in other logging.

test/Microsoft.AspNetCore.Antiforgery.Test/Internal/DefaultAntiforgeryTest.cs
```diff
@@ -225,7 +227,7 @@ public void GetAndStoreTokens_NoExistingCookieToken_Saved()
         }
 
         [Fact]
-        public async Task IsRequestValidAsync_FromStore_Failure()
+        public async Task IsRequestValueAsync_FromStore_Failure()
```
dougbu: Don't revert this fix!!! Desperately important.

dougbu:
Shouldn't `IsRequestValidAsync()` and the other `public` `IAntiforgery` methods (which use almost entirely separate code paths), especially `GetAndStoreTokens()`, also log?

src/Microsoft.AspNetCore.Antiforgery/Internal/DefaultAntiforgeryLoggerExtensions.cs
```diff
@@ -0,0 +1,77 @@
...
+            _validationFailed = LoggerMessage.Define<string>(
+                LogLevel.Warning,
+                1,
+                "Antiforgery validation failed with this message '{Message}'");
+            _validating = LoggerMessage.Define(
+                LogLevel.Information,
+                2,
+                "Antiforgery is attempting to validate a request.");
+            _missingCookieToken = LoggerMessage.Define<string>(
+                LogLevel.Warning,
```
dougbu: Wondering whether this should be a less-frequent message e.g. `LogLevel.Information`. Every user comes to a site initially with no cookie token and many will never log in.

dougbu:
:watch:

src/Microsoft.AspNetCore.Antiforgery/Internal/DefaultAntiforgery.cs
Eilon: I think this code is perfectly clear as it is.

src/Microsoft.AspNetCore.Antiforgery/Internal/DefaultAntiforgeryLoggerExtensions.cs
Eilon: Is that when this message is logged? It looks to me that this is logged when a form is submitted without the required cookie - that's an application error. This is different from just plain old boring anonymous stuff happening, no?

dougbu: This is called if `ValidateRequestAsync()` is called and the cookie token isn't present. I agree we're not going to call this method for GET requests, but can we say the same of user code?

src/Microsoft.AspNetCore.Antiforgery/Internal/DefaultAntiforgery.cs
```diff
@@ -102,11 +106,14 @@ public AntiforgeryTokenSet GetTokens(HttpContext httpContext)
                 throw new ArgumentNullException(nameof(httpContext));
             }
 
+            _logger.ValidatingAntiforgeryToken();
+
             CheckSSLConfig(httpContext);
```
dougbu: This method will `throw` if it's an HTTP request and Antiforgery is configured to require HTTPS. Should that event be logged?

src/Microsoft.AspNetCore.Antiforgery/Internal/DefaultAntiforgeryLoggerExtensions.cs
Eilon: It's going to throw an exception 1 statement later, why wouldn't this always be a log warning?

dougbu: So we should use a different logging level for the same event in `IsRequestValidAsync()`? If yes, sounds good to me.

src/Microsoft.AspNetCore.Antiforgery/Internal/DefaultAntiforgery.cs
rynowak: No, the exception will be logged where it's caught.

src/Microsoft.AspNetCore.Antiforgery/Internal/DefaultAntiforgery.cs
rynowak: I'd skip the 'validating' logging, and just log the result.

src/Microsoft.AspNetCore.Antiforgery/Internal/DefaultAntiforgeryLoggerExtensions.cs
rynowak: If you're recalling a discussion between lou and I, no that was about something related but not this.

test/Microsoft.AspNetCore.Antiforgery.Test/Internal/DefaultAntiforgeryTest.cs
ryanbrandenburg: Huh, must have messed up the rebase, will revert.

ryanbrandenburg:
Is there a specific thing we want to track with the other public IAntiforgery methods? They're not used by any of our code, existing for other implementations of IAntiforgery. It doesn't really seem like there's much to say about them other than "this was called" which I'm not sure is worth logging.

src/Microsoft.AspNetCore.Antiforgery/Internal/DefaultAntiforgery.cs
ryanbrandenburg: :see_no_evil:

test/Microsoft.AspNetCore.Antiforgery.Test/Internal/DefaultAntiforgeryTest.cs
ryanbrandenburg: I assume this is :see_no_evil:, don't know why GitHub trimmed off the "old" line

src/Microsoft.AspNetCore.Antiforgery/Internal/DefaultAntiforgeryLoggerExtensions.cs
ryanbrandenburg: Sounds like :see_no_evil: then.

src/Microsoft.AspNetCore.Antiforgery/Internal/DefaultAntiforgery.cs
ryanbrandenburg: I went ahead with @rynowak 's suggestion.

ryanbrandenburg:
:up: :date:

dougbu:
> They're not used by any of our code

MVC uses `SetCookieTokenAndHeader()` as well as `GetAndStoreTokens()`. In addition `ValidateAntiforgeryTokenAuthorizationFilter` (if it needs to exist still, I'm not sure) should really call `IsRequestValidAsync()` and not `ValidateRequestAsync()` (in a `try` block).

ryanbrandenburg:
:up: :date:

src/Microsoft.AspNetCore.Antiforgery/Internal/DefaultAntiforgeryLoggerExtensions.cs
```diff
@@ -0,0 +1,77 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using Microsoft.Extensions.Logging;
+
+namespace Microsoft.AspNetCore.Antiforgery.Internal
+{
+    internal static class DefaultAntiforgeryLoggerExtensions
```
rynowak: You can just call this `AntiforgeryLoggerExtensions`. If we add more logging to the package, it will likely go here as well.

rynowak:
:shipit: from me

dougbu:
:shipit: from me 2

PR #54, opened 2016-02-12
 * Add logging for other IAntiforgery public methods

### LINKED ISSUE #44: Add logging to antiforgery

ryanbrandenburg:
Fixes #44 
cc @dougbu and @rynowak

### DIFF
--- a/src/Microsoft.AspNetCore.Antiforgery/Internal/DefaultAntiforgery.cs
+++ b/src/Microsoft.AspNetCore.Antiforgery/Internal/DefaultAntiforgery.cs
@@ -59,7 +59,6 @@ namespace Microsoft.AspNetCore.Antiforgery.Internal
 
                     SaveCookieTokenAndHeader(httpContext, antiforgeryContext.NewCookieTokenString);
                     antiforgeryContext.HaveStoredNewCookieToken = true;
-
                     _logger.NewCookieToken();
                 }
                 else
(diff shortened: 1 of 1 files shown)


src/Microsoft.AspNetCore.Antiforgery/Internal/DefaultAntiforgery.cs
```diff
@@ -301,6 +295,15 @@ private AntiforgeryContext GetCookieTokens(HttpContext httpContext)
...
 
+            if (antiforgeryContext.HaveGeneratedNewCookieToken)
+            {
+                _logger.NewCookieToken();
+            }
+            else
+            {
+                _logger.ReusedCookieToken();
+            }
+
```
dougbu: Remove this addition.

Would potentially logging multiple times per request. Should instead restore the block where `SaveCookieTokenAndHeader()` is called in `GetAndStoreTokens()`. Then duplicate that logic in `SetCookieTokenAndHeader()`. That way, logging can reliably be done once -- when `HaveStoredNewCookieToken` is `false`.

Note as well that this method is called in cases (`GetTokens()`) where the cookie isn't stored at all.

dougbu:
:watch:

ryanbrandenburg:
:up: :date: @dougbu I believe that's what you meant?

dougbu:
@ryanbrandenburg yup, this is what I meant.

dougbu:
:shipit:

PR #55, opened 2016-02-12
Make IsRequestValid check HTTP method

rynowak:
This code was popping up everywhere this method is called. Seems bad to
duplicate it. Really what the caller wants to know is 'is the request
valid or a potential CSRF exploit?'. This gets the API closer to that.

### DIFF
--- a/src/Microsoft.AspNetCore.Antiforgery/IAntiforgery.cs
+++ b/src/Microsoft.AspNetCore.Antiforgery/IAntiforgery.cs
@@ -36,12 +36,13 @@ namespace Microsoft.AspNetCore.Antiforgery
         AntiforgeryTokenSet GetTokens(HttpContext httpContext);
 
         /// <summary>
-        /// Asynchronously returns a value indicating whether the request contains a valid antiforgery token.
+        /// Asynchronously returns a value indicating whether the request passes antiforgery validation. If the
+        /// request uses a safe HTTP method (GET, HEAD, OPTIONS, TRACE), the antiforgery token is not validated.
         /// </summary>
         /// <param name="httpContext">The <see cref="HttpContext"/> associated with the current request.</param>
         /// <returns>
-        /// A <see cref="Task{bool}"/> that, when completed, returns <c>true</c> if the request contains a
-        /// valid antiforgery token, otherwise returns <c>false</c>.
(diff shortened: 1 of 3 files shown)


src/Microsoft.AspNetCore.Antiforgery/IAntiforgery.cs
```diff
@@ -36,12 +36,13 @@ public interface IAntiforgery
...
-        /// Asynchronously returns a value indicating whether the request contains a valid antiforgery token.
+        /// Asynchronously returns a value indicating whether the request passes antiforgery validation. If the
+        /// request uses a safe HTTP method (GET, HEAD, OPTIONS, TRACE), the antiforgery token is not validated.
         /// </summary>
         /// <param name="httpContext">The <see cref="HttpContext"/> associated with the current request.</param>
         /// <returns>
-        /// A <see cref="Task{bool}"/> that, when completed, returns <c>true</c> if the request contains a
-        /// valid antiforgery token, otherwise returns <c>false</c>.
+        /// A <see cref="Task{bool}"/> that, when completed, returns <c>true</c> if the is valid, otherwise 
+        /// returns <c>false</c>.
```
dougbu: Hard to determine from these comments what the method will return if the request uses a safe HTTP method. Does "not validated" mean it returns `true` or `false`? Please clarify.

test/Microsoft.AspNetCore.Antiforgery.Test/Internal/DefaultAntiforgeryTest.cs
```diff
@@ -364,6 +365,74 @@ public void ValidateTokens_MissingCookieInTokenSet_Throws()
...
+            var result = await antiforgery.IsRequestValidAsync(context.HttpContext);
+
+            // Assert
+            Assert.True(result);
+            context.TokenGenerator
+                .Verify(o => o.TryValidateTokenSet(
+                    context.HttpContext,
+                    context.TestTokenSet.OldCookieToken,
+                    context.TestTokenSet.RequestToken,
+                    out message), Times.Never);
```
dougbu: new line for `Times.Never` :trollface:

dougbu:
:watch: just to see the updated doc comments

Issue #23, opened 2015-11-21
Improve the performance of Antiforgery


rynowak:
Right now the cost of using `@Html.Antiforgery` is about 7.5kb / request. In the context of a small MVC form, this is about 15% of the total bytes. We can do much better.

Some places to start looking.

Strings:
![image](https://cloud.githubusercontent.com/assets/1430011/11316049/71f184c6-8fb2-11e5-9ac2-b95493e37d4b.png)

Bytes:
![image](https://cloud.githubusercontent.com/assets/1430011/11316055/84dac3c2-8fb2-11e5-8163-f9ec2c2ad58a.png)

dougbu:
@rynowak this report duplicates existing issues such as #3123 since `new HtmlString()` is used to make things worse in `@Html.Antiforgery()`. I'll ignore anti-forgery over the weekend. But let's discuss possible overlaps on Monday.

rynowak:
Let's talk about avoiding overlap then :+1: I didn't realize part of this was already tracked.

The byte[] stuff AFAIK is hitherto untracked work.

Eilon:
@dougbu please continue to work with @rynowak on how to approach this.

dougbu:
- 492c079
- 0ddfa5f
- At least one more bit coming after extending the `WebEncoders.Base64UrlEncode()` / `WebEncoders.Base64UrlDecode()` API.

rynowak:
We should also verify that we're not paying the cost of re-serializing the cookie token when a request provides one. Here's the scenario:

**Request 1** GET /Products/Edit/5
Client: no tokens provided
Server: sets cookie token and renders request token in the form

**Request 2** POST /Products/Edit/5
Client: provides cookie token and request token
Server: 
    - deserializes the cookie and request token for validation
    - server does not need to set the cookie as it's already set
    - server does a redirect to /Products

Our memory profiling doesn't capture this scenario because we're not using cookies.

dougbu:
FYI @rynowak I'm doing the check you suggest. Suspect all is well because `tokenSet.IsNewCookieToken` handling at least appears correct.

dougbu:
[...]

PR #56, opened 2016-02-16
Pool `char`s used for base64url-encoding and -decoding

dougbu:
- #23 part 4
- depends on aspnet/HttpAbstractions#563

nits:
- correct name of a field in `AntiforgerySerializationContext`
- avoid allocations when returning an `AntiforgerySerializationContext` in (unlikely) case `Stream` is unused

### DIFF
--- a/src/Microsoft.AspNetCore.Antiforgery/Internal/DefaultAntiforgeryTokenSerializer.cs
+++ b/src/Microsoft.AspNetCore.Antiforgery/Internal/DefaultAntiforgeryTokenSerializer.cs
@@ -45,12 +45,16 @@ namespace Microsoft.AspNetCore.Antiforgery.Internal
                 var count = serializedToken.Length;
                 var charsRequired = WebEncoders.GetArraySizeRequiredToDecode(count);
                 var chars = serializationContext.GetChars(charsRequired);
-                serializedToken.CopyTo(0, chars, 0, count);
-                var tokenBytes = WebEncoders.Base64UrlDecode(chars, 0, count);
+                var tokenBytes = WebEncoders.Base64UrlDecode(
+                    serializedToken,
+                    offset: 0,
+                    count: count,
+                    buffer: chars,
+                    bufferOffset: 0);
 
                 var unprotectedBytes = _cryptoSystem.Unprotect(tokenBytes);
                 var stream = serializationContext.Stream;
-                stream.Write(unprotectedBytes, 0, unprotectedBytes.Length);
+                stream.Write(unprotectedBytes, offset: 0, count: unprotectedBytes.Length);
                 stream.Position = 0L;
 
                 var reader = serializationContext.Reader;
@@ -166,9 +170,14 @@ namespace Microsoft.AspNetCore.Antiforgery.Internal
                 var count = bytes.Length;
                 var charsRequired = WebEncoders.GetArraySizeRequiredToEncode(count);
                 var chars = serializationContext.GetChars(charsRequired);
-                var outputLength = WebEncoders.Base64UrlEncode(bytes, 0, count, chars, 0);
+                var outputLength = WebEncoders.Base64UrlEncode(
+                    bytes,
+                    offset: 0,
+                    count: count,
+                    output: chars,
+                    outputOffset: 0);
 
-                return new string(chars, 0, outputLength);
+                return new string(chars, startIndex: 0, length: outputLength);
             }
             finally
             {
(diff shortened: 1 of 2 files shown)


src/Microsoft.AspNetCore.Antiforgery/Internal/DefaultAntiforgeryTokenSerializer.cs
```diff
@@ -156,7 +161,14 @@ public string Serialize(AntiforgeryToken token)
...
                 var stream = serializationContext.Stream;
-                return WebEncoders.Base64UrlEncode(_cryptoSystem.Protect(stream.ToArray()));
+                var bytes = _cryptoSystem.Protect(stream.ToArray());
+
+                var count = bytes.Length;
+                var charsRequired = WebEncoders.GetArraySizeRequiredToEncode(count);
+                var chars = serializationContext.GetChars(charsRequired);
+                var outputLength = WebEncoders.Base64UrlEncode(bytes, 0, count, chars, 0);
+
+                return new string(chars, 0, outputLength);
```
dougbu: Can't avoid this allocation w/o changing the properties in `AntiforgeryTokenSet` from `string` to `ArraySegment<char>` or similar. Then would need `serializationContext.GetChars1()` and `serializationContext.GetChars2()` for the two tokens. Ick all 'round.

src/Microsoft.AspNetCore.Antiforgery/Internal/DefaultAntiforgeryTokenSerializer.cs
```diff
@@ -156,7 +161,14 @@ public string Serialize(AntiforgeryToken token)
 
                 writer.Flush();
                 var stream = serializationContext.Stream;
-                return WebEncoders.Base64UrlEncode(_cryptoSystem.Protect(stream.ToArray()));
+                var bytes = _cryptoSystem.Protect(stream.ToArray());
```
dougbu: Can't avoid the `ToArray()` allocation even after working around the need to `stream.GetBuffer()` for .NET 4.51 and `stream.TryGetBuffer()` otherwise. Like the `byte[]` allocation within `Protect()`, this is fallout from the `byte[]` to `byte[]` `IDataProtector` API.

dougbu:
###### Before
pooling no

###### After
pooling yes

src/Microsoft.AspNetCore.Antiforgery/Internal/DefaultAntiforgeryTokenSerializer.cs
rynowak: The way to do this would be to:
1). Use `new MemoryString(byte[])` 
2). Implement a custom memory stream based on pooled memory

src/Microsoft.AspNetCore.Antiforgery/Internal/DefaultAntiforgeryTokenSerializer.cs
rynowak: What's the impact of this allocation specifically? should we fix it? (not this PR)

rynowak:
:shipit:

src/Microsoft.AspNetCore.Antiforgery/Internal/DefaultAntiforgeryTokenSerializer.cs
dougbu: First, no, I don't think we should work to avoid this allocation. We'd have a hideous public API if we used almost anything but `string`s.

Second, the impact is a bit under 2 MB (~1.4% of total allocations) in the test scenario:
after strings

src/Microsoft.AspNetCore.Antiforgery/Internal/DefaultAntiforgeryTokenSerializer.cs
dougbu: Because the `IDataProtector.Protect()` argument is a `byte[]` of the correct length and not an `ArraySegment<byte>`, @rynowak's approach would work only if we

1. Calculated the `MemoryStream` buffer size in advance.
2. Serialized a lot of tokens with _exactly_ the same initial serialization size.
3. Pooled / reused a few `byte[]` instances of different sizes.
4. Managed contention between requests.

dougbu: BTW this allocation alone is not huge, likely because it occurs prior to encryption and URL-encoding, both of which expand. Just a bit more than a quarter the `byte`s allocated in this `Protect()` call for example:
after bytes

dougbu: Hmm, the `ToArray()` allocates about 13% as much as the `new string(...)` a few lines later.

dougbu:
Thx for the sign-off @rynowak. I'd like your perspective on aspnet/HttpAbstractions#563, especially @Tratcher's [recent comment](https://github.com/aspnet/HttpAbstractions/pull/563#discussion-diff-53090645).

dougbu:
:up::date:

dougbu:
478edc1

PR #58, opened 2016-02-18
Add antiforgery middleware

rynowak:
This new middleware participates in authentication and acts as a filter
when the request doesn't include a valid CSRF token for a POST (or other unsafe verb).

Any authentication middleware that you want to validate an antiforgery
token should go ahead of this middleware in the pipeline (Cookies,
IISIntegration). This also takes care of automatic auth (Windows) done by
weblistener.

Any authentication middleware that you want to ignore antiforgery should
go after this middleware in the pipeline.

To facilitate this, there are a few changes in the antiforgery API
surface. Namely we can now pass in a principal to validate tokens. You
can't pass in a principal to generate tokens - we expect you to be logged
in at that poing. Also, ValidateRequestAsync(...) now checks the HTTP verb
and won't validate GETs and such.

### DIFF
--- a/src/Microsoft.AspNetCore.Antiforgery/AntiforgeryBuilderExtensions.cs
+++ b/src/Microsoft.AspNetCore.Antiforgery/AntiforgeryBuilderExtensions.cs
@@ -1,0 +1,30 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using Microsoft.AspNetCore.Antiforgery;
+
+namespace Microsoft.AspNetCore.Builder
+{
+    /// <summary>
+    /// Extension methods for configuring the antiforgery middleware.
+    /// </summary>
+    public static class AntiforgeryBuilderExtensions
+    {
+        /// <summary>
+        /// Adds the antiforgery middleware.
+        /// </summary>
+        /// <param name="app">The <see cref="IApplicationBuilder"/>.</param>
+        /// <returns>The <see cref="IApplicationBuilder"/>.</returns>
+        public static IApplicationBuilder UseAntiforgery(this IApplicationBuilder app)
+        {
+            if (app == null)
+            {
+                throw new ArgumentNullException(nameof(app));
+            }
+
+
+            return app.UseMiddleware<AntiforgeryMiddleware>();
+        }
+    }
+}
--- a/src/Microsoft.AspNetCore.Antiforgery/AntiforgeryBuilderExtensions.cs
+++ b/src/Microsoft.AspNetCore.Antiforgery/AntiforgeryBuilderExtensions.cs
@@ -23,7 +23,6 @@ namespace Microsoft.AspNetCore.Builder
                 throw new ArgumentNullException(nameof(app));
             }
 
-
             return app.UseMiddleware<AntiforgeryMiddleware>();
         }
     }
--- a/src/Microsoft.AspNetCore.Antiforgery/AntiforgeryMiddleware.cs
+++ b/src/Microsoft.AspNetCore.Antiforgery/AntiforgeryMiddleware.cs
@@ -1,0 +1,70 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.Antiforgery.Internal;
+using Microsoft.AspNetCore.Http;
+
+namespace Microsoft.AspNetCore.Antiforgery
+{
+    /// <summary>
+    /// A middleware implementation of antiforgery validation.
+    /// </summary>
+    public class AntiforgeryMiddleware
+    {
+        /// <summary>
+        /// Creates a new <see cref="AntiforgeryMiddleware"/>.
+        /// </summary>
+        /// <param name="next">The <see cref="RequestDelegate"/> for the next middleware.</param>
+        /// <param name="antiforgery">The <see cref="IAntiforgery"/>.</param>
+        public AntiforgeryMiddleware(RequestDelegate next, IAntiforgery antiforgery)
+        {
+            if (next == null)
+            {
+                throw new ArgumentNullException(nameof(next));
+            }
+
+            if (antiforgery == null)
+            {
+                throw new ArgumentNullException(nameof(antiforgery));
+            }
+
+            Next = next;
+            Antiforgery = antiforgery;
+        }
+
+        /// <summary>
+        /// Gets the <see cref="IAntiforgery"/>.
+        /// </summary>
+        protected IAntiforgery Antiforgery { get; }
+
+        /// <summary>
+        /// Gets the <see cref="RequestDelegate"/> for the next middleware.
+        /// </summary>
+        protected RequestDelegate Next { get; }
+
+        /// <summary>
+        /// Invokes the middleware for the given <paramref name="httpContext"/>.
+        /// </summary>
+        /// <param name="httpContext">The <see cref="HttpContext"/> associated with the current request.</param>
+        /// <returns>A <see cref="Task"/> which will be completed when execution of the middleware completes.</returns>
+        public async Task Invoke(HttpContext httpContext)
+        {
+            if (httpContext == null)
+            {
+                throw new ArgumentNullException(nameof(httpContext));
+            }
+
+            var handler = CreateHandler();
+            await handler.InitializeAsync(httpContext);
+
+            await Next(httpContext);
+        }
+
+        private AntiforgeryAuthenticationHandler CreateHandler()
+        {
+            return new AntiforgeryAuthenticationHandler(Antiforgery);
+        }
+    }
+}
--- a/src/Microsoft.AspNetCore.Antiforgery/AntiforgeryMiddleware.cs
+++ b/src/Microsoft.AspNetCore.Antiforgery/AntiforgeryMiddleware.cs
@@ -13,6 +13,9 @@ namespace Microsoft.AspNetCore.Antiforgery
     /// </summary>
     public class AntiforgeryMiddleware
     {
+        private readonly IAntiforgery _antiforgery;
+        private readonly RequestDelegate _next;
+
         /// <summary>
         /// Creates a new <see cref="AntiforgeryMiddleware"/>.
         /// </summary>
@@ -30,19 +33,9 @@ namespace Microsoft.AspNetCore.Antiforgery
                 throw new ArgumentNullException(nameof(antiforgery));
             }
 
-            Next = next;
-            Antiforgery = antiforgery;
+            _next = next;
+            _antiforgery = antiforgery;
         }
-
-        /// <summary>
-        /// Gets the <see cref="IAntiforgery"/>.
-        /// </summary>
-        protected IAntiforgery Antiforgery { get; }
-
-        /// <summary>
-        /// Gets the <see cref="RequestDelegate"/> for the next middleware.
-        /// </summary>
-        protected RequestDelegate Next { get; }
 
         /// <summary>
         /// Invokes the middleware for the given <paramref name="httpContext"/>.
@@ -56,15 +49,17 @@ namespace Microsoft.AspNetCore.Antiforgery
                 throw new ArgumentNullException(nameof(httpContext));
             }
 
-            var handler = CreateHandler();
+            var handler = new AntiforgeryAuthenticationHandler(_antiforgery);
             await handler.InitializeAsync(httpContext);
 
-            await Next(httpContext);
-        }
-
-        private AntiforgeryAuthenticationHandler CreateHandler()
-        {
-            return new AntiforgeryAuthenticationHandler(Antiforgery);
+            try
+            {
+                await _next(httpContext);
+            }
+            finally
+            {
+                handler.Teardown();
+            }
         }
     }
 }
--- a/src/Microsoft.AspNetCore.Antiforgery/IAntiforgery.cs
+++ b/src/Microsoft.AspNetCore.Antiforgery/IAntiforgery.cs
@@ -1,6 +1,7 @@
 // Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+using System.Security.Claims;
 using System.Threading.Tasks;
 using Microsoft.AspNetCore.Http;
 
@@ -54,6 +55,16 @@ namespace Microsoft.AspNetCore.Antiforgery
         /// Thrown when the request does not include a valid antiforgery token.
         /// </exception>
         Task ValidateRequestAsync(HttpContext httpContext);
+
+        /// <summary>
+        /// Validates an antiforgery token that was supplied as part of the request.
+        /// </summary>
+        /// <param name="httpContext">The <see cref="HttpContext"/> associated with the current request.</param>
+        /// <param name="principal">The claims-based principal to validate.</param>
+        /// <exception cref="AntiforgeryValidationException">
+        /// Thrown when the request does not include a valid antiforgery token.
+        /// </exception>
+        Task ValidateRequestAsync(HttpContext httpContext, ClaimsPrincipal principal);
 
         /// <summary>
         /// Generates and stores an antiforgery cookie token if one is not available or not valid.
--- a/src/Microsoft.AspNetCore.Antiforgery/IAntiforgery.cs
+++ b/src/Microsoft.AspNetCore.Antiforgery/IAntiforgery.cs
@@ -48,6 +48,18 @@ namespace Microsoft.AspNetCore.Antiforgery
         Task<bool> IsRequestValidAsync(HttpContext httpContext);
 
         /// <summary>
+        /// Asynchronously returns a value indicating whether the request passes antiforgery validation. If the
+        /// request uses a safe HTTP method (GET, HEAD, OPTIONS, TRACE), the antiforgery token is not validated.
+        /// </summary>
+        /// <param name="httpContext">The <see cref="HttpContext"/> associated with the current request.</param>
+        /// <param name="principal">The claims-based principal to validate.</param>
+        /// <returns>
+        /// A <see cref="Task{Boolean}"/> that, when completed, returns <c>true</c> if the is requst uses a safe HTTP
+        /// method or contains a value antiforgery token, otherwise returns <c>false</c>.
+        /// </returns>
+        Task<bool> IsRequestValidAsync(HttpContext httpContext, ClaimsPrincipal principal);
+
+        /// <summary>
         /// Validates an antiforgery token that was supplied as part of the request.
         /// </summary>
         /// <param name="httpContext">The <see cref="HttpContext"/> associated with the current request.</param>
--- a/src/Microsoft.AspNetCore.Antiforgery/Internal/AntiforgeryAuthenticationHandler.cs
+++ b/src/Microsoft.AspNetCore.Antiforgery/Internal/AntiforgeryAuthenticationHandler.cs
@@ -48,21 +48,19 @@ namespace Microsoft.AspNetCore.Antiforgery.Internal
             if (PriorHandler != null)
             {
                 await PriorHandler.AuthenticateAsync(context);
-            }
 
-            var authentication = GetAuthenticationFeature(HttpContext);
-            var oldPrincipal = authentication.User;
-
-            if (context.Principal != null)
-            {
-                try
-                { 
-                    await Antiforgery.ValidateRequestAsync(HttpContext, context.Principal);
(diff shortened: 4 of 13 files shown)


rynowak:
Currently design-tagged as I'm still writing tests for the middleware and auth handler.

src/Microsoft.AspNetCore.Antiforgery/IAntiforgery.cs
```diff
@@ -56,6 +57,16 @@ public interface IAntiforgery
...
 
         /// <summary>
+        /// Validates an antiforgery token that was supplied as part of the request.
+        /// </summary>
+        /// <param name="httpContext">The <see cref="HttpContext"/> associated with the current request.</param>
+        /// <param name="principal">The claims-based principal to validate.</param>
+        /// <exception cref="AntiforgeryValidationException">
+        /// Thrown when the request does not include a valid antiforgery token.
+        /// </exception>
+        Task ValidateRequestAsync(HttpContext httpContext, ClaimsPrincipal principal);
```
rynowak: @dougbu - should we add the same for `IsRequestValidAsync`? I don't need it for what I'm doing but it would be consistent.

src/Microsoft.AspNetCore.Antiforgery/Internal/IAntiforgeryTokenGenerator.cs
```diff
@@ -20,9 +21,13 @@ public interface IAntiforgeryTokenGenerator
         /// Generates a request token corresponding to <paramref name="cookieToken"/>.
         /// </summary>
         /// <param name="httpContext">The <see cref="HttpContext"/> associated with the current request.</param>
+        /// <param name="principal">The claims-based principal to use for token generation.</param>
         /// <param name="cookieToken">A valid cookie token.</param>
         /// <returns>An <see cref="AntiforgeryToken"/>.</returns>
-        AntiforgeryToken GenerateRequestToken(HttpContext httpContext, AntiforgeryToken cookieToken);
+        AntiforgeryToken GenerateRequestToken(
+            HttpContext httpContext,
+            ClaimsPrincipal principal,
```
rynowak: Not strictly required, but it's nice to look at the code and see that `context.User` isn't used anywhere.

src/Microsoft.AspNetCore.Antiforgery/project.json
```diff
@@ -14,6 +14,7 @@
   "dependencies": {
     "Microsoft.AspNetCore.DataProtection": "1.0.0-*",
     "Microsoft.AspNetCore.Http.Abstractions": "1.0.0-*",
+    "Microsoft.AspNetCore.Http": "1.0.0-*",
```
rynowak: Needed for `HttpAuthenticationFeature`

Tratcher:
I'm still a little concerned about the ordering requirement in practice. It would be easy for people to become vulnerable by forgetting this middleware or putting it too early in the pipeline.

src/Microsoft.AspNetCore.Antiforgery/AntiforgeryMiddleware.cs
```diff
@@ -0,0 +1,70 @@
...
+            }
+
+            Next = next;
+            Antiforgery = antiforgery;
+        }
+
+        /// <summary>
+        /// Gets the <see cref="IAntiforgery"/>.
+        /// </summary>
+        protected IAntiforgery Antiforgery { get; }
```
Tratcher: protected?

rynowak: What's the suggestion? `private`?

rynowak:
> I'm still a little concerned about the ordering requirement in practice. It would be easy for people to become vulnerable by forgetting this middleware or putting it too early in the pipeline.

I'm not going to try to say that's not a thing, but this seems like the best of all of our options. 

/cc @blowdart

blowdart:
So yea ordering sucks. But if we get it right in the templates that goes a long way to illustrating how to use it.

This is a general middleware problem though. If only we could check that other middlewares are already in place first.

src/Microsoft.AspNetCore.Antiforgery/AntiforgeryMiddleware.cs
Tratcher: Yes, private until proven otherwise.

Tratcher: or just private fields.

src/Microsoft.AspNetCore.Antiforgery/IAntiforgery.cs
HaoK: Would the principal usually be context.User?  If not, when is it different

src/Microsoft.AspNetCore.Antiforgery/AntiforgeryMiddleware.cs
```diff
@@ -0,0 +1,70 @@
...
+        /// <returns>A <see cref="Task"/> which will be completed when execution of the middleware completes.</returns>
+        public async Task Invoke(HttpContext httpContext)
+        {
+            if (httpContext == null)
+            {
+                throw new ArgumentNullException(nameof(httpContext));
+            }
+
+            var handler = CreateHandler();
+            await handler.InitializeAsync(httpContext);
```
Tratcher: Auth middleware should also detach their handler after Next finishes.

src/Microsoft.AspNetCore.Antiforgery/IAntiforgery.cs
rynowak: https://github.com/aspnet/Antiforgery/pull/58/files#diff-07e8828d724692198761a77330ce9fc3R60

we need to pass in `AuthenticateContext.User`

src/Microsoft.AspNetCore.Antiforgery/Internal/AntiforgeryAuthenticationHandler.cs
```diff
@@ -0,0 +1,124 @@
...
+
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.Antiforgery.Internal;
+using Microsoft.AspNetCore.Http;
+using Microsoft.AspNetCore.Http.Features.Authentication;
+using Microsoft.AspNetCore.Http.Features.Authentication.Internal;
+
+namespace Microsoft.AspNetCore.Antiforgery.Internal
+{
+    public class AntiforgeryAuthenticationHandler : IAuthenticationHandler
```
HaoK: Now that we have two of these, I almost think we should push most of this code into a WrappingAuthenticationHandler base class, so this and claims transformation only have to override the methods they care about, and a lot of this mechanical goop can go away.

src/Microsoft.AspNetCore.Antiforgery/AntiforgeryMiddleware.cs
Tratcher: In a finally block

src/Microsoft.AspNetCore.Antiforgery/IAntiforgery.cs
HaoK: Ok, makes sense, you are validating the principal that's being authenticated, not the principal that's associated with the request...  Although this principal may eventually be set on the context...

src/Microsoft.AspNetCore.Antiforgery/AntiforgeryMiddleware.cs
rynowak: :+1:

src/Microsoft.AspNetCore.Antiforgery/IAntiforgery.cs
dougbu: > should we add the same for `IsRequestValidAsync`?

I vote yea, verily.

src/Microsoft.AspNetCore.Antiforgery/AntiforgeryBuilderExtensions.cs
```diff
@@ -0,0 +1,30 @@
...
+        /// <param name="app">The <see cref="IApplicationBuilder"/>.</param>
+        /// <returns>The <see cref="IApplicationBuilder"/>.</returns>
+        public static IApplicationBuilder UseAntiforgery(this IApplicationBuilder app)
+        {
+            if (app == null)
+            {
+                throw new ArgumentNullException(nameof(app));
+            }
+
+
```
dougbu: Did you leave something out? :trollface:

src/Microsoft.AspNetCore.Antiforgery/Internal/AntiforgeryAuthenticationHandler.cs
```diff
@@ -0,0 +1,124 @@
...
+                }
+            }
+        }
+
+        /// <inheritdoc />
+        public async Task AuthenticateAsync(AuthenticateContext context)
+        {
+            if (PriorHandler != null)
+            {
+                await PriorHandler.AuthenticateAsync(context);
```
Tratcher: It's actually an error condition for the prior handler to be null, no?

rynowak: What if another middleware does automatic auth only (no handler)? I thought about this case but it's not clear to me that it's 100% an error condition.

src/Microsoft.AspNetCore.Antiforgery/AntiforgeryMiddleware.cs
rynowak: BTW this isn't the case for **the** authN middleware https://github.com/aspnet/Security/blob/dev/src/Microsoft.AspNetCore.Authentication/AuthenticationMiddleware.cs - it does it in a `catch`

Should I open a bug for this?

src/Microsoft.AspNetCore.Antiforgery/Internal/AntiforgeryAuthenticationHandler.cs
rynowak: Also, what about weblistener + nothing for windows auth. That's a valid scenario where there'd be no prior handler.

src/Microsoft.AspNetCore.Antiforgery/AntiforgeryMiddleware.cs
```diff
@@ -0,0 +1,70 @@
...
+
+            await Next(httpContext);
+        }
+
+        private AntiforgeryAuthenticationHandler CreateHandler()
```
dougbu: Curious: Why not inline this one-liner?

src/Microsoft.AspNetCore.Antiforgery/Internal/IAntiforgeryTokenGenerator.cs
dougbu: Lots of hmming and hawing over here. Do see the slight cleanliness of not seeing `context.User` used. But I'm not sure it makes sense for this method to support as bizarre a scenario as generating a token for anyone but the current user. Can imagine future developers looking at this and trying to figure out a realistic scenario where passing both `httpContext` and `principal` makes sense.

test/Microsoft.AspNetCore.Antiforgery.Test/Internal/DefaultAntiforgeryTest.cs
```diff
@@ -800,6 +829,135 @@ public void GetAndStoreTokens_DoesNotStoreTwice()
         }
 
         [Theory]
+        [InlineData("GeT")]
+        [InlineData("HEAD")]
+        [InlineData("options")]
+        [InlineData("TrAcE")]
```
dougbu: Think I've seen this data set somewhere before :smiley_cat:

src/Microsoft.AspNetCore.Antiforgery/Internal/AntiforgeryAuthenticationHandler.cs
```diff
@@ -0,0 +1,124 @@
...
+using Microsoft.AspNetCore.Antiforgery.Internal;
+using Microsoft.AspNetCore.Http;
+using Microsoft.AspNetCore.Http.Features.Authentication;
+using Microsoft.AspNetCore.Http.Features.Authentication.Internal;
+
+namespace Microsoft.AspNetCore.Antiforgery.Internal
+{
+    public class AntiforgeryAuthenticationHandler : IAuthenticationHandler
+    {
+        public AntiforgeryAuthenticationHandler(IAntiforgery antiforgery)
```
dougbu: Where / when will this guy or `AntiforgeryMiddleware` be tested?

rynowak: > [Design]...

src/Microsoft.AspNetCore.Antiforgery/Internal/AntiforgeryAuthenticationHandler.cs
```diff
@@ -0,0 +1,124 @@
...
+        {
+            Antiforgery = antiforgery;
+        }
+
+        protected IAntiforgery Antiforgery { get; }
```
dougbu: Similar to @Tratcher's [comment](https://github.com/aspnet/Antiforgery/pull/58/files#r53392610), `protected` seems weird in a `class` with no `virtual` methods.

src/Microsoft.AspNetCore.Antiforgery/Internal/AntiforgeryAuthenticationHandler.cs
```diff
@@ -0,0 +1,124 @@
...
+
+        public HttpContext HttpContext { get; private set; }
+
+        public IAuthenticationHandler PriorHandler { get; private set; }
```
dougbu: `IAuthenticationHandler` doesn't require these. So why are they `public`?

src/Microsoft.AspNetCore.Antiforgery/AntiforgeryMiddleware.cs
Tratcher: Yes they do in TeardownAsync

src/Microsoft.AspNetCore.Antiforgery/Internal/AntiforgeryAuthenticationHandler.cs
Tratcher: Don't make @rynowak do that. You can try to consolidate later.

src/Microsoft.AspNetCore.Antiforgery/Internal/AntiforgeryAuthenticationHandler.cs
Tratcher: WebListener does register as an auth handler so it can do customized challenges. So does platform handler.

src/Microsoft.AspNetCore.Antiforgery/Internal/IAntiforgeryTokenGenerator.cs
Tratcher: When doing passive auth HttpContext.User may never be set, you may always retrieve it by calling HttpContext.Authentication.AuthenticateAsync

src/Microsoft.AspNetCore.Antiforgery/project.json
Tratcher: You can remove Abstractions and WebUtilities then.

src/Microsoft.AspNetCore.Antiforgery/AntiforgeryMiddleware.cs
rynowak: `TeardownAsync` is only called in a `catch` - and never on success unless I'm missing something

Tratcher: Hmm, why isn't that a finally....https://github.com/aspnet/Security/issues/704

rynowak:
Added tests for the middleware/handler - removed design tag. Next will be addressing feedback.

src/Microsoft.AspNetCore.Antiforgery/IAntiforgery.cs
rynowak: Added

src/Microsoft.AspNetCore.Antiforgery/project.json
rynowak: :+1:

src/Microsoft.AspNetCore.Antiforgery/Internal/AntiforgeryAuthenticationHandler.cs
rynowak: I haven't made any changes here yet. If you're really convinced that this is important and have a suggestion about a sensible thing for us to do, please make a suggestion :laughing:

rynowak:
@dougbu @Tratcher @HaoK updated.

I think I've addressed all the actionable feedback - except [this](https://github.com/aspnet/Antiforgery/pull/58#discussion_r53494040)

src/Microsoft.AspNetCore.Antiforgery/Internal/AntiforgeryAuthenticationHandler.cs
Tratcher: Nevermind. If the middleware is placed out of order this may never even be called because the upstream provider would short circuit:
https://github.com/aspnet/Security/blob/dev/src/Microsoft.AspNetCore.Authentication/AuthenticationHandler.cs#L200-L233

src/Microsoft.AspNetCore.Antiforgery/Internal/AntiforgeryAuthenticationHandler.cs
```diff
@@ -0,0 +1,161 @@
...
+            if (context == null)
+            {
+                throw new ArgumentNullException(nameof(context));
+            }
+
+            if (_priorHandler != null)
+            {
+                await _priorHandler.AuthenticateAsync(context);
+
+                var authentication = GetAuthenticationFeature(_httpContext);
```
Tratcher: This line looks redundant.

test/Microsoft.AspNetCore.Antiforgery.Test/AntiforgeryMiddlewareTest.cs
```diff
@@ -0,0 +1,382 @@
...
+            services.AddLogging();
+            services.AddOptions();
+
+            var antiforgery = new Mock<IAntiforgery>(MockBehavior.Strict);
+            services.AddSingleton(antiforgery.Object);
+
+            var result = new TestContext();
+            result.Antiforgery = antiforgery;
+
+            var app = new ApplicationBuilder(services.BuildServiceProvider());
```
Tratcher: Use TestServer rather than setting up this whole thing by yourself

Tratcher:
Cleanup and :shipit:

src/Microsoft.AspNetCore.Antiforgery/Internal/DefaultAntiforgery.cs
```diff
@@ -268,6 +310,21 @@ private void CheckSSLConfig(HttpContext context)
...
 
+        private bool IsValidationRequired(HttpContext httpContext)
+        {
+            var method = httpContext.Request.Method;
+            if (string.Equals(method, "GET", StringComparison.OrdinalIgnoreCase) ||
+                string.Equals(method, "HEAD", StringComparison.OrdinalIgnoreCase) ||
+                string.Equals(method, "OPTIONS", StringComparison.OrdinalIgnoreCase) ||
+                string.Equals(method, "TRACE", StringComparison.OrdinalIgnoreCase))
+            {
+                // Validation not needed for these request types.
```
javiercn: I don't feel this comment adds anything

src/Microsoft.AspNetCore.Antiforgery/Internal/IAntiforgeryTokenGenerator.cs
javiercn: :+1: I believe this is also how it happens in OWIN scenarios. I feel it's better to always pass in the user principal

src/Microsoft.AspNetCore.Antiforgery/Internal/DefaultAntiforgery.cs
rynowak: I'll touch it up a bit.

src/Microsoft.AspNetCore.Antiforgery/Internal/AntiforgeryAuthenticationHandler.cs
rynowak: Yup :+1:

test/Microsoft.AspNetCore.Antiforgery.Test/AntiforgeryMiddlewareTest.cs
rynowak: How does that expose the app func? This is already really simple

PR #59, opened 2016-02-22
Updated antiforgery ServiceCollectionExtensions

javiercn:
@Eilon @kichalla Sample pattern to follow. Docs are based on the MVC version

### DIFF
--- a/src/Microsoft.AspNetCore.Antiforgery/ServiceCollectionExtensions.cs
+++ b/src/Microsoft.AspNetCore.Antiforgery/ServiceCollectionExtensions.cs
@@ -10,9 +10,16 @@ using Microsoft.Extensions.Options;
 
 namespace Microsoft.Extensions.DependencyInjection
 {
+    /// <summary>
+    /// Extension methods for setting up antiforgery services in an <see cref="IServiceCollection" />.
+    /// </summary>
     public static class ServiceCollectionExtensions
     {
-        public static IServiceCollection AddAntiforgery(this IServiceCollection services)
+        /// <summary>
+        /// Adds antiforgery services to the specified <see cref="IServiceCollection" />.
+        /// </summary>
+        /// <param name="services">The <see cref="IServiceCollection" /> to add services to.</param>
+        public static void AddAntiforgery(this IServiceCollection services)
         {
             if (services == null)
             {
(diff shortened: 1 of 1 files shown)


src/Microsoft.AspNetCore.Antiforgery/ServiceCollectionExtensions.cs
```diff
@@ -40,11 +47,14 @@ public static IServiceCollection AddAntiforgery(this IServiceCollection services
...
 
-        public static IServiceCollection AddAntiforgery(
+        /// <summary>
+        /// Adds antiforgery services to the specified <see cref="IServiceCollection" />.
+        /// </summary>
+        /// <param name="services">The <see cref="IServiceCollection" /> to add services to.</param>
+        /// <param name="setupAction">An action delegate to configure the provided <see cref="AntiforgeryOptions"/>.</param>
+        public static void AddAntiforgery(
             this IServiceCollection services,
             Action<AntiforgeryOptions> setupAction)
```
kichalla: Add a null check for the delegate.

javiercn:
Things to pay attention to:
* Comments (There's been a lot of copy paste, so mistakes can happen).
* Method names
* Null checks for arguments
* Overload with configuration action calls parameterless overload.
* Parameterless overload does the work and brings in direct depenendencies
  (Like routing on MVC, for most repositories it means making sure AddOptions is called).

src/Microsoft.AspNetCore.Antiforgery/ServiceCollectionExtensions.cs
```diff
@@ -10,9 +10,16 @@
...
 namespace Microsoft.Extensions.DependencyInjection
 {
+    /// <summary>
+    /// Extension methods for setting up antiforgery services in an <see cref="IServiceCollection" />.
+    /// </summary>
     public static class ServiceCollectionExtensions
     {
-        public static IServiceCollection AddAntiforgery(this IServiceCollection services)
+        /// <summary>
+        /// Adds antiforgery services to the specified <see cref="IServiceCollection" />.
```
pranavkm: `specified <paramref name="services" />`?

javiercn: It doesn't make sense, services in that context is not referring to the services parameter

javiercn: Forget my comment, I don't have a strong take between using paramref vs see in this case, is there a clear advantage on doing that? I'm saying it because it's not trivial to go through all the repositories changing this.

dougbu: Consider it forgotten. I prefer `<paramref/>` because that's more consistent with `<summary>`s throughout MVC. I also think making the comments word-for-word identical across our repos should be a non-goal.

Eilon: I think the comment is fine as it is. I think any reader would perfectly well understand what this does.

Eilon:
:shipit:

src/Microsoft.AspNetCore.Antiforgery/ServiceCollectionExtensions.cs
dougbu: After [discussing](https://github.com/aspnet/Mvc/pull/4159#discussion-diff-54208453) this point in aspnet/Mvc#4159, I agree completely. @javiercn should leave this as-is.

PR #65, opened 2016-03-07
Fix package metadata

Eilon:
@rynowak


src/Microsoft.AspNetCore.Antiforgery/project.json
```diff
@@ -1,5 +1,9 @@
 {
-  "description": "An antiforgery system for ASP.NET designed to generate and validate tokens to prevent Cross-Site Request Forgery attacks.",
+  "description": "An antiforgery system for ASP.NET Core designed to generate and validate tokens to prevent Cross-Site Request Forgery attacks.",
+  "tags": [
+    "aspnetcore",
+    "antiforgery"
```
dougbu: Also tag the frequent over-hyphentation -- "anti-forgery"?

Eilon: That'd be wrong, so meh :smile:

PR #78, opened 2016-04-18
Preserve X-Frame-Options if it was already set

### LINKED ISSUE #60: DefaultAntiforgery.SaveCookieTokenAndHeader should not override stricter X-Frame-Options headers
Currently, `SaveCookieTokenAndHeader` always replaces the `X-Frame-Options` header with its own value, even if a stricter header has already been added to the response headers by another middleware (e.g NWebsec).

Of course, `SuppressXFrameOptionsHeader` could be used to prevent `SaveCookieTokenAndHeader` from updating the `X-Frame-Options` header but I believe it would be better to add a check to avoid overriding it in this case, specially if the existing header is stricter (i.e uses DENY).

/cc @blowdart

ryanbrandenburg:
Fixes #60.

As per @rynowak and @Eilon's comments, we're now preserving the `X-Frame-Options` header in the response if it's already been set.


ryanbrandenburg:
CC @dougbu

dougbu:
:shipit:

rynowak:
:shipit:

PR #80, opened 2016-04-25
Update web.config and add publish tool

BrennanConroy:
@JunTaoLuo @Tratcher @Eilon

### DIFF
--- a/samples/AntiforgerySample/project.json
+++ b/samples/AntiforgerySample/project.json
@@ -30,5 +30,14 @@
   "content": [
     "wwwroot",
     "web.config"
-  ]
+  ],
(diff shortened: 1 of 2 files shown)


samples/AntiforgerySample/project.json
```diff
@@ -30,5 +30,14 @@
   "content": [
     "wwwroot",
```
moozzyk: is there any content in wwwroot that needs to be copied or this line can be removed too?

BrennanConroy: some js files and index.html

moozzyk:
:ship::it:

PR #81, opened 2016-05-19
Change priority for request token source lookup. Header token now tak…

kichalla:
…es priority over form field token.

@rynowak @Tratcher


test/Microsoft.AspNetCore.Antiforgery.Test/Internal/DefaultAntiforgeryTokenStoreTest.cs
```diff
@@ -231,35 +236,6 @@ public void GetCookieToken_CookieIsNotEmpty_ReturnsToken()
             Assert.Null(tokenSet.RequestToken);
         }
 
-        [Fact]
-        public async Task GetFormToken_FormFieldIsValid_ReturnsToken()
```
rynowak: Why was this test removed? Is it a dupe of one of the others?

rynowak:
:shipit: once we resolve what's going on with that test

PR #92, opened 2016-06-19
Use Angular lib instead of angular map file in sample

ciwchris:
A reference to angular did not exist and so the sample app would error with a message to that effect.


Eilon:
@ryanbrandenburg can you review and merge?

ryanbrandenburg:
Thanks @ciwchris! Merging.

Issue #95, opened 2016-07-28
How do I modify the defaults for Antiforgery to make it FIPS-compliant


thanhos:
I'm working on a system that has FIPS-compliant enabled and am wondering is there a way to change the Antiforgery defaults to make it compliant.  If someone could provide some direction or feedback that would appreciated.

Thanks

thanhos:
Creating a new ASP.NET Core Web Application using .NET Core works fine.  Creating one using the .NET Framework with FIPS enabled doesn't work.

blowdart:
Strange.

What OS, and do you have a stack trace?

@rynowak This may end up being deep in data protection. There's a fun bug in creating SHA signing pieces where you tell it to use FIPS algorithms and it doesn't. nuget had the same problem,

thanhos:
I'm on Windows 10.0 (Build 10240).

Here's the stack trace:

fail: Microsoft.AspNetCore.Diagnostics.DeveloperExceptionPageMiddleware[0]
      An unhandled exception has occurred while executing the request
System.Reflection.TargetInvocationException: Exception has been thrown by the target of an invocation. ---> System.InvalidOperationException: This implementation is not part of the Windows Platform FIPS validated cryptographic algorithms.
   at System.Security.Cryptography.SHA256Managed..ctor()
   --- End of inner exception stack trace ---
   at System.RuntimeMethodHandle.InvokeMethod(Object target, Object[] arguments, Signature sig, Boolean constructor)
   at System.Reflection.RuntimeConstructorInfo.Invoke(BindingFlags invokeAttr, Binder binder, Object[] parameters, CultureInfo culture)
   at System.Security.Cryptography.CryptoConfig.CreateFromName(String name, Object[] args)
   at System.Security.Cryptography.SHA256.Create()
   at Microsoft.AspNetCore.Antiforgery.Internal.AntiforgeryOptionsSetup.ComputeCookieName(String applicationId)
   at Microsoft.AspNetCore.Antiforgery.Internal.AntiforgeryOptionsSetup.ConfigureOptions(AntiforgeryOptions options, DataProtectionOptions dataProtectionOptions)
   at Microsoft.Extensions.Options.OptionsCache`1.CreateOptions()
   at System.Threading.LazyInitializer.EnsureInitializedCore[T](T& target, Boolean& initialized, Object& syncLock, Func`1 valueFactory)
   at Microsoft.Extensions.Options.OptionsCache`1.get_Value()
[...]

PR #96, opened 2016-08-02
Change SHA256 algorithm to work on FIPS-compliant machines.

NTaylorMullen:
#95

### DIFF
--- a/src/Microsoft.AspNetCore.Antiforgery/Internal/AntiforgeryOptionsSetup.cs
+++ b/src/Microsoft.AspNetCore.Antiforgery/Internal/AntiforgeryOptionsSetup.cs
@@ -28,7 +28,7 @@ namespace Microsoft.AspNetCore.Antiforgery.Internal
 
         private static string ComputeCookieName(string applicationId)
         {
-            using (var sha256 = SHA256.Create())
+            using (var sha256 = CryptographyAlgorithms.CreateSHA256())
             {
                 var hash = sha256.ComputeHash(Encoding.UTF8.GetBytes(applicationId));
                 var subHash = hash.Take(8).ToArray();
--- a/src/Microsoft.AspNetCore.Antiforgery/Internal/AntiforgerySerializationContext.cs
+++ b/src/Microsoft.AspNetCore.Antiforgery/Internal/AntiforgerySerializationContext.cs
@@ -88,7 +88,7 @@ namespace Microsoft.AspNetCore.Antiforgery.Internal
             {
                 if (_sha256 == null)
                 {
-                    _sha256 = SHA256.Create();
+                    _sha256 = CryptographyAlgorithms.CreateSHA256();
                 }
 
                 return _sha256;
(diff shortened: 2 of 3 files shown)


src/Microsoft.AspNetCore.Antiforgery/Internal/AntiforgeryOptionsSetup.cs
```diff
@@ -28,7 +28,14 @@ public static void ConfigureOptions(AntiforgeryOptions options, DataProtectionOp
 
         private static string ComputeCookieName(string applicationId)
         {
-            using (var sha256 = SHA256.Create())
+            HashAlgorithm sha256;
+#if NETSTANDARD1_3
+            sha256 = SHA256.Create();
+#else
+            sha256 = new SHA256Cng();
```
PinpointTownes: This will crash on Mono: https://github.com/aspnet/Antiforgery/issues/95#issuecomment-236988168

I know Mono is not an "officially supported" target, but I'm not sure adding more pain is the right thing to do.

PinpointTownes: After verification, it looks like the type exists on Mono. It just uses `SHA256Managed` internally.

PinpointTownes:
It's important to note that your "fix" has a side effect: one won't be able to override the implementation used for SHA-256 using `CryptoConfig.AddAlgorithm`, since you'll manually instantiate it on .NET Desktop.

Kinda sucks, IMHO.

src/Microsoft.AspNetCore.Antiforgery/Internal/AntiforgeryOptionsSetup.cs
dougbu: Wasn't `(HashAlgorithm)new SHA256CryptoServiceProvider()` recommended in one of the (zillion) #95 comments? Not sure of the advantages / disadvantages but wanted to check.

NTaylorMullen: @PinpointTownes Thanks for the feedback! @blowdart is currently on vacation but once he returns I will run it across him.

@dougbu: May have been, this approach is what @blowdart recommended offline.

blowdart: Mono compatibility is not a goal.

NTaylorMullen: @blowdart does the desktop `SHA256Cng` implementation on Windows pull from the registered cryptography algorithms? I'm more concerned with disabling the ability to hot swap crypto algorithms.

NTaylorMullen:
:up: :date: 

This new approach should still enable users to override the crypto algorithm (as long as its fips compliant).

PinpointTownes:
This is a much better approach :+1: 

Any particular reason to use CAPI (`SHA256CryptoServiceProvider`) instead of the newer CNG-based API (`SHA256Cng`)?

NTaylorMullen:
@PinpointTownes Only reason was consistency. We don't use `SHA256Cng` in our codebases but we do use `SHA256CryptoServiceProvider`.

Eilon:
:shipit:

PinpointTownes:
> @PinpointTownes Only reason was consistency. 

Consistency always wins :+1:

blowdart:
:shipit:

PR #107, opened 2016-10-28
[Fixes #105] Disable caching when response uses antiforgery

### LINKED ISSUE #105: Disable caching when response uses antiforgery
When setting the cookie token and headers on a response that uses antiforgery, caching should be disabled to prevent the content from being cached.

Adding the cache control headers `Cache-Control: no-cache, no-store` should be sufficient.

cc @sebastienros @Tratcher

kichalla:
@rynowak 

Added functional tests in MVC as part of this PR: https://github.com/aspnet/Mvc/pull/5468

### DIFF
--- a/src/Microsoft.AspNetCore.Antiforgery/IAntiforgery.cs
+++ b/src/Microsoft.AspNetCore.Antiforgery/IAntiforgery.cs
@@ -14,7 +14,8 @@ namespace Microsoft.AspNetCore.Antiforgery
     {
         /// <summary>
         /// Generates an <see cref="AntiforgeryTokenSet"/> for this request and stores the cookie token
-        /// in the response.
+        /// in the response. This operation also sets the "Cache-control" and "Pragma" headers to "no-cache" and
+        /// the "X-Frame-Options" header to "SAMEORIGIN".
         /// </summary>
         /// <param name="httpContext">The <see cref="HttpContext"/> associated with the current request.</param>
         /// <returns>An <see cref="AntiforgeryTokenSet" /> with tokens for the response.</returns>
--- a/src/Microsoft.AspNetCore.Antiforgery/Internal/AntiforgeryLoggerExtensions.cs
+++ b/src/Microsoft.AspNetCore.Antiforgery/Internal/AntiforgeryLoggerExtensions.cs
@@ -15,6 +15,7 @@ namespace Microsoft.AspNetCore.Antiforgery.Internal
         private static readonly Action<ILogger, Exception> _newCookieToken;
         private static readonly Action<ILogger, Exception> _reusedCookieToken;
         private static readonly Action<ILogger, Exception> _tokenDeserializeException;
+        private static readonly Action<ILogger, Exception> _responseCacheHeadersOverridenToNoCache;
 
         static AntiforgeryLoggerExtensions()
         {
@@ -47,6 +48,11 @@ namespace Microsoft.AspNetCore.Antiforgery.Internal
                 LogLevel.Error,
                 7,
                 "An exception was thrown while deserializing the token.");
+            _responseCacheHeadersOverridenToNoCache = LoggerMessage.Define(
+                LogLevel.Warning,
+                8,
+                "The 'Cache-Control' and 'Pragma' headers have been overridden and set to 'no-cache' to prevent " +
+                "caching of this response. Any response that uses antiforgery should not be cached.");
         }
 
         public static void ValidationFailed(this ILogger logger, string message)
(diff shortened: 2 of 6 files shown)


src/Microsoft.AspNetCore.Antiforgery/IAntiforgery.cs
```diff
@@ -14,7 +14,7 @@ public interface IAntiforgery
     {
         /// <summary>
         /// Generates an <see cref="AntiforgeryTokenSet"/> for this request and stores the cookie token
-        /// in the response.
+        /// in the response. It also sets the "Cache-control" and "Pragma" headers to "no-cache".
```
rynowak: Replace "It also" with "This operation also". I'd also mention the `X-Frame-Options` as well here if you're going to describe the header actions.

src/Microsoft.AspNetCore.Antiforgery/Internal/AntiforgeryLoggerExtensions.cs
```diff
@@ -47,6 +48,10 @@ static AntiforgeryLoggerExtensions()
                 LogLevel.Error,
                 7,
                 "An exception was thrown while deserializing the token.");
+            _responseNotCached = LoggerMessage.Define(
+                LogLevel.Debug,
+                8,
+                "The 'Cache-Control' and 'Pragma' headers of the response have been set as 'no-cache'.");
```
rynowak: Rather than doing this, I think it would be much better to check if the `Cache-Control` and `Pragma` headers were already set and log a **warning** log message if they are.

test/Microsoft.AspNetCore.Antiforgery.Test/Internal/DefaultAntiforgeryTest.cs
```diff
@@ -276,6 +277,67 @@ public void GetAndStoreTokens_ExistingValidCookieToken_NotOverriden()
         }
 
         [Fact]
+        public void GetAndStoreTokens_ExistingValidCookieToken_NotOverriden_AndSetsDoNotCacheHeaders()
```
rynowak: Is this really the right behaviour? Should this be different from `SetCookieTokenAndHeader`?

kichalla: The behavior here is similar to `SetCookieTokenAndHeader`. Not sure If I confused you with the method name here but cache headers are always overriden...the `NotOverriden` part here was for the cookie token.

rynowak: OK, I see now, I misread the name earlier

src/Microsoft.AspNetCore.Antiforgery/Internal/DefaultAntiforgery.cs
```diff
@@ -358,6 +367,43 @@ private IAntiforgeryFeature GetTokensInternal(HttpContext httpContext)
...
+
+            httpContext.Response.Headers[HeaderNames.CacheControl] = "no-cache";
+            httpContext.Response.Headers[HeaderNames.Pragma] = "no-cache";
+        }
+
+        private void LogCacheHeaderOverrideWarning(HttpResponse response)
+        {
+            var logWarning = false;
+            CacheControlHeaderValue cacheControlHeaderValue;
+            if (CacheControlHeaderValue.TryParse(response.Headers[HeaderNames.CacheControl], out cacheControlHeaderValue))
```
kichalla: Not sure if this parsing is a perf concern as this method would be called in a typical view scenario. Let me know if I you need to me to add a check in the beginning to see if the logger is enabled to begin with.

kichalla:
🆙 📅

src/Microsoft.AspNetCore.Antiforgery/Internal/DefaultAntiforgery.cs
kichalla: Ignore my previous comment. I do not see any adverse affect due to this change based on perf data.

test/Microsoft.AspNetCore.Antiforgery.Test/Internal/DefaultAntiforgeryTest.cs
```diff
@@ -2,19 +2,25 @@
...
+using Microsoft.Net.Http.Headers;
 using Moq;
 using Xunit;
 
 namespace Microsoft.AspNetCore.Antiforgery.Internal
 {
     public class DefaultAntiforgeryTest
     {
+        private const string ResponseCacheHeadersOverrideWarningMessage =
+            "Response's cache headers have been overriden and set to 'no-cache'.";
```
rynowak: The 'Cache-Control' and 'Pragma' headers have been overridden and set to 'no-cache' to prevent caching of this response. Any response that uses antiforgery should not be cached.

rynowak: This makes it clear what headers were changed and why

PR #108, opened 2016-10-31
[Fixes #104] Expose cookie options via Antiforgery options

### LINKED ISSUE #104: Expose cookie options via Antiforgery options
With the following commit, the path of the antiforgery's cookie token is set to always use the request's `PathBase`. Consider exposing the cookie's options via the `AntiforgeryOptions` to enable users to set it explicitly if they need to.

kichalla:


src/Microsoft.AspNetCore.Antiforgery/AntiforgeryOptions.cs
```diff
@@ -46,6 +47,16 @@ public string CookieName
         }
 
         /// <summary>
+        /// The path set on the cookie. Defaults to current request's <see cref="HttpRequest.PathBase"/> value.
+        /// </summary>
+        public PathString? CookiePath { get; set; }
+
+        /// <summary>
+        /// The domain set on the cookie. Defaults to <c>null</c>.
```
rynowak: Yeah, but what does it **mean** for it to be null? That would be useful to know

src/Microsoft.AspNetCore.Antiforgery/AntiforgeryOptions.cs
```diff
@@ -46,6 +47,16 @@ public string CookieName
         }
 
         /// <summary>
+        /// The path set on the cookie. Defaults to current request's <see cref="HttpRequest.PathBase"/> value.
```
rynowak: I would say that - if it's null, we use the current request's pathbase. It's not that **this** property will be initialized to the pathbase

PR #109, opened 2016-11-08
Fix antiforgery cookie from being set to empty pathbase

kichalla:
@Eilon @pranavkm @dougbu 

This is a similar issue as @DamianEdwards found here: https://github.com/aspnet/Mvc/issues/5512

### DIFF
--- a/test/Microsoft.AspNetCore.Antiforgery.FunctionalTests/AntiforgerySampleTest.cs
+++ b/test/Microsoft.AspNetCore.Antiforgery.FunctionalTests/AntiforgerySampleTest.cs
@@ -6,6 +6,7 @@ using System.Linq;
 using System.Net;
 using System.Net.Http;
 using System.Threading.Tasks;
+using Microsoft.Net.Http.Headers;
 using Xunit;
 
 namespace Microsoft.AspNetCore.Antiforgery.FunctionalTests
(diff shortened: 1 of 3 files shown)


test/Microsoft.AspNetCore.Antiforgery.FunctionalTests/AntiforgerySampleTest.cs
```diff
@@ -26,11 +27,19 @@ public AntiforgerySampleTests(AntiForgerySampleTestFixture fixture)
...
 
             // Assert
-            var cookie = RetrieveAntiforgeryCookie(response);
-            Assert.NotNull(cookie.Value);
-
-            var token = RetrieveAntiforgeryToken(response);
-            Assert.NotNull(token.Value);
+            var setCookieHeaderValue = RetrieveAntiforgeryCookie(response);
+            Assert.NotNull(setCookieHeaderValue);
+            Assert.False(string.IsNullOrEmpty(setCookieHeaderValue.Value));
```
Eilon: Isn't there some specific assertion for this?

test/Microsoft.AspNetCore.Antiforgery.FunctionalTests/AntiforgerySampleTest.cs
```diff
@@ -26,11 +27,19 @@ public AntiforgerySampleTests(AntiForgerySampleTestFixture fixture)
...
+            Assert.Null(setCookieHeaderValue.Domain);
+            Assert.Equal("/", setCookieHeaderValue.Path);
+            Assert.False(setCookieHeaderValue.Secure);
+
+            setCookieHeaderValue = RetrieveAntiforgeryToken(response);
```
Eilon: Are these lines identical to the block above?

test/Microsoft.AspNetCore.Antiforgery.FunctionalTests/AntiforgerySampleTest.cs
kichalla: I do not see a specific assertion for this...we can do in 2 steps..check for notnull and assert.notempty...

test/Microsoft.AspNetCore.Antiforgery.FunctionalTests/AntiforgerySampleTest.cs
kichalla: Yes, I thought of moving to a common method, but I think in general we want to avoid these common methods as it immediately does not reflect the line where the test fails. I am not too strong about it though.

test/Microsoft.AspNetCore.Antiforgery.FunctionalTests/AntiforgerySampleTest.cs
Eilon: Oh no nevermind then.

test/Microsoft.AspNetCore.Antiforgery.FunctionalTests/AntiforgerySampleTest.cs
Eilon: No, I mean aren't lines 30-35 identical to 37-42?

Eilon: Ohhhhhhhhhh my eyes have failed me. Nevermind, then!

PR #110, opened 2016-11-09
Use request PathBase value to set cookie path only if it has a non-nu…

kichalla:
…ll & non-empty value

@Eilon @DamianEdwards

### DIFF
--- a/src/Microsoft.AspNetCore.Antiforgery/AntiforgeryOptions.cs
+++ b/src/Microsoft.AspNetCore.Antiforgery/AntiforgeryOptions.cs
@@ -47,8 +47,9 @@ namespace Microsoft.AspNetCore.Antiforgery
         }
 
         /// <summary>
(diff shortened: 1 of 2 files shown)


src/Microsoft.AspNetCore.Antiforgery/AntiforgeryOptions.cs
```diff
@@ -48,7 +48,8 @@ public string CookieName
 
         /// <summary>
         /// The path set on the cookie. If it's <c>null</c>, the "path" attribute on the cookie is set to current
-        /// request's <see cref="HttpRequest.PathBase"/> value.
+        /// request's <see cref="HttpRequest.PathBase"/> value. If the value of <see cref="HttpRequest.PathBase"/> is
```
Eilon: Do this: https://github.com/aspnet/Mvc/pull/5521#discussion_r87125822

Then :shipit:

PR #117, opened 2017-01-23
[Fixes #116] Set 'no-store' also in Cache-Conrol header

### LINKED ISSUE #116: Question: `no-store` versus `no-cache` for Antiforgery
This question is related to how caching and Antiforgery was handled for https://github.com/aspnet/Antiforgery/issues/105.

Looking at https://github.com/aspnet/Antiforgery/blob/3fc090e2fee47884af7968031e40876fcb6549ae/src/Microsoft.AspNetCore.Antiforgery/Internal/DefaultAntiforgery.cs#L70-L72 ...

> Explicitly set the cache headers to 'no-cache'. This could override any user set value but this is fine as a response with antiforgery token must never be cached.

and thus `SetDoNotCacheHeaders()` ...
https://github.com/aspnet/Antiforgery/blob/3fc090e2fee47884af7968031e40876fcb6549ae/src/Microsoft.AspNetCore.Antiforgery/Internal/DefaultAntiforgery.cs#L370-L379

@JunTaoLuo said [ref](https://github.com/aspnet/Antiforgery/issues/105#issuecomment-252769270) ...
> The proposal is to disable caching when antiforgery is used regardless of whether cookies are being set. 

[RFC 7234](https://tools.ietf.org/html/rfc7234) doesn't guarantee that storage will not occur tho ...

> 5.2.1.4.  no-cache
> The "no-cache" request directive indicates that a cache MUST NOT use a stored response to satisfy the request without successful validation on the origin server.

Wondering: Why doesn't Antiforgery set `no-store` rather than `no-cache`?

kichalla:


src/Microsoft.AspNetCore.Antiforgery/Internal/DefaultAntiforgery.cs
```diff
@@ -367,14 +365,18 @@ private IAntiforgeryFeature GetTokensInternal(HttpContext httpContext)
             return antiforgeryFeature;
         }
 
-        private void SetDoNotCacheHeaders(HttpContext httpContext)
+        /// <summary>
+        /// Sets the 'Cache-Control' header to 'no-cache, no-store' and 'Pragma' header to 'no-cache' overriding any user set value.
+        /// </summary>
+        /// <param name="httpContext">The <see cref="HttpContext"/>.</param>
+        protected virtual void SetDoNotCacheHeaders(HttpContext httpContext)
```
kichalla: Looking at the following issue, it appears there are multiple values/headers that users have seen work for their scenarios. 
http://stackoverflow.com/questions/49547/how-to-control-web-page-caching-across-all-browsers

src/Microsoft.AspNetCore.Antiforgery/Internal/DefaultAntiforgery.cs
```diff
@@ -367,14 +365,18 @@ private IAntiforgeryFeature GetTokensInternal(HttpContext httpContext)
...
         {
             // Since antifogery token generation is not very obvious to the end users (ex: MVC's form tag generates them
             // by default), log a warning to let users know of the change in behavior to any cache headers they might
             // have set explicitly.
             LogCacheHeaderOverrideWarning(httpContext.Response);
 
-            httpContext.Response.Headers[HeaderNames.CacheControl] = "no-cache";
+            httpContext.Response.Headers[HeaderNames.CacheControl] = "no-cache, no-store";
```
JunTaoLuo: An alternative here would be to check if these directives already exist and append them if not present. This preserves other directives like `must-revalidate` and possible cache extensions. I'm satisfied with this approach since it looks like most clients will not cache entries with both `no-cache` and `no-store` but maybe someone should take a closer look.

rynowak: I'd prefer that we wait for feedback on this issue as that would requires us to understand the cache headers in AF. For instance we've generally always overwritten what's there with `no-cache` and no one has complained

JunTaoLuo: Well we don't need to understand the cache headers, you can potentially use something like https://github.com/aspnet/HttpAbstractions/blob/dev/src/Microsoft.Net.Http.Headers/HeaderUtilities.cs#L289. But i'm okay with the current approach as well.

src/Microsoft.AspNetCore.Antiforgery/Internal/DefaultAntiforgery.cs
JunTaoLuo: Yes, but `no-cache, no-store` should be enough for most scenarios. We can add more or change this mechanism if people ask for it.

PR #126, opened 2017-03-01
Fixed typo in IAntiforgery file

jnaputi253:
Removed the `is` word, and changed `requst` to `request` on line 44.


jnaputi253:
I have a quick question: on line 45, it reads

> /// method or contains a value antiforgery token, otherwise returns <c>false</c>.

Should it read

> /// method or contains an antiforgery token, otherwise returns <c>false</c>.

instead?

Eilon:
@kichalla can you take a look?

kichalla:
Probably instead of "method or contains a value antiforgery token" it should be "method or contains a valid antiforgery token"?

jnaputi253:
@kichalla Based on the rest of the documentation, I can agree with that.  I have adjusted the sentence accordingly.

If there is a disagreement with the grammar, please say so.

kichalla:
1c9a4f5705c02d182e6beefa96739de81cd3d850

Thanks! @jnaputi253

PR #140, opened 2017-05-18
Update TFM to netstandard2.0

natemcmaster:


build/common.props
```diff
@@ -16,7 +16,7 @@
     <PackageReference Include="Internal.AspNetCore.Sdk" Version="$(InternalAspNetCoreSdkVersion)" PrivateAssets="All" />
   </ItemGroup>
 
-  <ItemGroup Condition="'$(TargetFrameworkIdentifier)'=='.NETFramework' AND '$(OutputType)'=='library'">
-    <PackageReference Include="NETStandard.Library" Version="$(BundledNETStandardPackageVersion)" />
+  <ItemGroup Condition="'$(TargetFrameworkIdentifier)'=='.NETFramework'">
```
JunTaoLuo: This seems better than the `Condition="'$(TargetFramework)'=='net461'"` I was using as the condition.

build/dependencies.props
```diff
@@ -3,7 +3,7 @@
     <AspNetCoreVersion>2.0.0-*</AspNetCoreVersion>
     <InternalAspNetCoreSdkVersion>2.1.0-*</InternalAspNetCoreSdkVersion>
     <MoqVersion>4.7.1</MoqVersion>
-    <NETStandardImplicitPackageVersion>$(BundledNETStandardPackageVersion)</NETStandardImplicitPackageVersion>
+    <NETStandardLibraryNETFrameworkVersion>2.0.0-*</NETStandardLibraryNETFrameworkVersion>
```
JunTaoLuo: Do we no longer need `NETStandardImplicitPackageVersion`? What was this used for previously?

natemcmaster: Not needed here. We still need it in any repo that has netstandard1.x projects.

cref https://github.com/dotnet/sdk/blob/0fcbd7b7c8eff2c4d1ccbadc2219edab8ce9e9dd/src/Tasks/Microsoft.NET.Build.Tasks/build/Microsoft.NET.Sdk.DefaultItems.targets#L54-L63

Issue #127, opened 2017-03-02
Make antiforgery cookies enable SameSite by default


Eilon:
See https://github.com/aspnet/Security/issues/908

PR #141, opened 2017-05-22
React to SameSite addition to CookieOptions

JunTaoLuo:
#127 We should be setting the default to strict?

### DIFF
--- a/src/Microsoft.AspNetCore.Antiforgery/AntiforgeryOptions.cs
+++ b/src/Microsoft.AspNetCore.Antiforgery/AntiforgeryOptions.cs
@@ -47,17 +47,29 @@ namespace Microsoft.AspNetCore.Antiforgery
         }
 
         /// <summary>
+        /// This is obsolete and will be removed in a future version.
+        /// The recommended alternative is to use ConfigureCookieOptions.
         /// The path set on the cookie. If set to <c>null</c>, the "path" attribute on the cookie is set to the current
         /// request's <see cref="HttpRequest.PathBase"/> value. If the value of <see cref="HttpRequest.PathBase"/> is
         /// <c>null</c> or empty, then the "path" attribute is set to the value of <see cref="CookieOptions.Path"/>.
         /// </summary>
+        [Obsolete("This is obsolete and will be removed in a future version. The recommended alternative is to use ConfigureCookieOptions.")]
         public PathString? CookiePath { get; set; }
 
         /// <summary>
-        /// The domain set on the cookie. By default its <c>null</c> which results in the "domain" attribute not being
-        /// set.
+        /// This is obsolete and will be removed in a future version.
+        /// The recommended alternative is to use ConfigureCookieOptions.
+        /// The domain set on the cookie. By default its <c>null</c> which results in the "domain" attribute not being set.
         /// </summary>
+        [Obsolete("This is obsolete and will be removed in a future version. The recommended alternative is to use ConfigureCookieOptions.")]
         public string CookieDomain { get; set; }
+
+        /// <summary>
+        /// Configures the <see cref="CookieOptions"/> of the antiforgery cookies. Without additional configuration, the 
+        /// default values antiforgery cookie options are true for <see cref="CookieOptions.HttpOnly"/>, null for
+        /// <see cref="CookieOptions.Domain"/> and <see cref="SameSiteMode.Strict"/> for <see cref="CookieOptions.SameSite"/>.
+        /// </summary>
+        public Action<HttpContext, CookieOptions> ConfigureCookieOptions { get; set; }
 
         /// <summary>
         /// Specifies the name of the antiforgery token field that is used by the antiforgery system.
--- a/src/Microsoft.AspNetCore.Antiforgery/Internal/DefaultAntiforgeryTokenStore.cs
+++ b/src/Microsoft.AspNetCore.Antiforgery/Internal/DefaultAntiforgeryTokenStore.cs
@@ -69,34 +69,34 @@ namespace Microsoft.AspNetCore.Antiforgery.Internal
             Debug.Assert(httpContext != null);
             Debug.Assert(token != null);
 
-            var options = new CookieOptions();
-            options.HttpOnly = true;
-            options.Domain = _options.CookieDomain;
-            // Note: don't use "newCookie.Secure = _options.RequireSSL;" since the default
-            // value of newCookie.Secure is populated out of band.
-            if (_options.RequireSsl)
+            var options = new CookieOptions
             {
-                options.Secure = true;
-            }
-            SetCookiePath(httpContext, options);
+                HttpOnly = true,
+#pragma warning disable 618
+                Domain = _options.CookieDomain,
+#pragma warning restore 618
+                SameSite = SameSiteMode.Strict,
+                Secure = _options.RequireSsl
+            };
 
-            httpContext.Response.Cookies.Append(_options.CookieName, token, options);
-        }
-
-        private void SetCookiePath(HttpContext httpContext, CookieOptions cookieOptions)
-        {
+#pragma warning disable 618
             if (_options.CookiePath != null)
             {
-                cookieOptions.Path = _options.CookiePath.ToString();
+                options.Path = _options.CookiePath.ToString();
             }
+#pragma warning restore 618
             else
             {
                 var pathBase = httpContext.Request.PathBase.ToString();
                 if (!string.IsNullOrEmpty(pathBase))
                 {
-                    cookieOptions.Path = pathBase;
+                    options.Path = pathBase;
                 }
             }
+
+            _options.ConfigureCookieOptions?.Invoke(httpContext, options);
+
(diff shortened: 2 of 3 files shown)


src/Microsoft.AspNetCore.Antiforgery/Internal/DefaultAntiforgeryTokenStore.cs
```diff
@@ -69,9 +69,13 @@ public void SaveCookieToken(HttpContext httpContext, string token)
...
             Debug.Assert(token != null);
 
-            var options = new CookieOptions();
-            options.HttpOnly = true;
-            options.Domain = _options.CookieDomain;
+            var options = new CookieOptions
+            {
+                HttpOnly = true,
+                Domain = _options.CookieDomain,
+                SameSite = SameSiteEnforcementMode.Strict
```
rynowak: I am not familiar with this option, Let's have a conversation with @blowdart - we need to understand what should be the default and **if** this should be configurable.

JunTaoLuo: Spoke offline, I think this should remain as strict as we expect the cookie to be created by requests from the same site as subsequent requests. There's no need to make this configurable.

JunTaoLuo:
Updated based on @blowdart's recommendation to make this option configurable. This is ready to merge.

rynowak:
If we're going to make this configurable we should attach an `Action<CookieOptions>` to `AntiforgeryOptions` and be done. Then users are empowered to do whatever they want and we don't end up duplicating all of `CookieOptions` on `AntiforgeryOptions`.

src/Microsoft.AspNetCore.Antiforgery/AntiforgeryOptions.cs
```diff
@@ -54,10 +54,11 @@ public string CookieName
...
 
         /// <summary>
-        /// The domain set on the cookie. By default its <c>null</c> which results in the "domain" attribute not being
-        /// set.
+        /// Configures the <see cref="CookieOptions"/> of the antiforgery cookies. Without additional configuration, the 
+        /// default values antiforgery cookie options are true for <see cref="CookieOptions.HttpOnly"/>, null for
+        /// <see cref="CookieOptions.Domain"/> and <see cref="SameSiteMode.Strict"/> for <see cref="CookieOptions.SameSite"/>.
         /// </summary>
-        public string CookieDomain { get; set; }
+        public Action<CookieOptions> ConfigureCookieOptions { get; set; }
```
JunTaoLuo: This is now a breaking change that requires an announcement.

src/Microsoft.AspNetCore.Antiforgery/Internal/DefaultAntiforgeryTokenStore.cs
```diff
@@ -80,6 +84,8 @@ public void SaveCookieToken(HttpContext httpContext, string token)
             }
             SetCookiePath(httpContext, options);
 
+            _options.ConfigureCookieOptions?.Invoke(options);
```
JunTaoLuo: I call the new ConfigureCookieOptions last after everything else is set. If we don't want users to be able to override Secure or Path on the CookieOptions, we can move this up, though I don't see why we might want that.

src/Microsoft.AspNetCore.Antiforgery/AntiforgeryOptions.cs
rynowak: Get @Eilon to sign off on the breaking change, I don't get to approve these

JunTaoLuo: Added him as a reviewer and will speak with him about the breaking change. Can you review the rest of the PR?

src/Microsoft.AspNetCore.Antiforgery/Internal/DefaultAntiforgeryTokenStore.cs
JunTaoLuo: Do we also want to configure the cookie path and secure in the delegate?

rynowak: Personally I think allowing you to set both is fine. RequireSSL is also enforced on the validation side, so if you mess it up, you'll just get a 400.

We should double check with @blowdart

test/Microsoft.AspNetCore.Antiforgery.Test/Internal/DefaultAntiforgeryTokenStoreTest.cs
```diff
@@ -364,7 +364,7 @@ public void SaveCookieToken_NonNullAntiforgeryOptionsCookieDomain_UsesOptionsCoo
                 .Returns("/index.html");
             var options = new AntiforgeryOptions();
             options.CookieName = _cookieName;
-            options.CookieDomain = expectedCookieDomain;
+            options.ConfigureCookieOptions = cookieOptions => cookieOptions.Domain = expectedCookieDomain;
```
rynowak: Change the name of this test to reflect the change.

src/Microsoft.AspNetCore.Antiforgery/Internal/DefaultAntiforgeryTokenStore.cs
blowdart: Consistency is good, but folks are already using the old way. I assume you don't want to remove the existing path, just add a way to set it here to?

JunTaoLuo: Yea, not changing the existing behaviour. Calling the configure delegate here allows the user to overwrite what we set but as @rynowak mentioned, they will be doing so at their own risk.

test/Microsoft.AspNetCore.Antiforgery.Test/Internal/DefaultAntiforgeryTokenStoreTest.cs
JunTaoLuo: Done

src/Microsoft.AspNetCore.Antiforgery/AntiforgeryOptions.cs
```diff
@@ -54,12 +54,21 @@ public string CookieName
         public PathString? CookiePath { get; set; }
```
rynowak: This should be obsolete too, it's redundant with ConfigureCookieOptions

src/Microsoft.AspNetCore.Antiforgery/Internal/DefaultAntiforgeryTokenStore.cs
```diff
@@ -69,9 +69,13 @@ public void SaveCookieToken(HttpContext httpContext, string token)
             Debug.Assert(httpContext != null);
             Debug.Assert(token != null);
 
-            var options = new CookieOptions();
-            options.HttpOnly = true;
-            options.Domain = _options.CookieDomain;
+            var options = new CookieOptions
+            {
+                HttpOnly = true,
+                Domain = null,
```
Tratcher: still use the obsolete domain property

src/Microsoft.AspNetCore.Antiforgery/Internal/DefaultAntiforgeryTokenStore.cs
```diff
@@ -69,9 +69,13 @@ public void SaveCookieToken(HttpContext httpContext, string token)
             Debug.Assert(httpContext != null);
             Debug.Assert(token != null);
 
-            var options = new CookieOptions();
-            options.HttpOnly = true;
-            options.Domain = _options.CookieDomain;
```
rynowak: We still need to use `_options.CookieDomain` - it's not being removed.

JunTaoLuo: ah yes

JunTaoLuo:
:up::date:

src/Microsoft.AspNetCore.Antiforgery/Internal/DefaultAntiforgeryTokenStore.cs
```diff
@@ -69,9 +69,13 @@ public void SaveCookieToken(HttpContext httpContext, string token)
...
+            var options = new CookieOptions
+            {
+                HttpOnly = true,
+                Domain = null,
+                SameSite = SameSiteMode.Strict
+            };
+
             // Note: don't use "newCookie.Secure = _options.RequireSSL;" since the default
             // value of newCookie.Secure is populated out of band.
             if (_options.RequireSsl)
```
Tratcher: This is misleading, remove it.

JunTaoLuo:
As per discussion with @Tratcher, we will probably adopt this approach for other components that duplicate settings on CookieOptions, potentially updating things like Session. Will file an issue in HttpAbstractions to track.

PR #148, opened 2017-06-29
Replace the ConfigureCookieOptions action property with the CookieBuilder

natemcmaster:
Addresses https://github.com/aspnet/HttpAbstractions/issues/853.

Uses the new CookieBuilder API added here: https://github.com/aspnet/HttpAbstractions/pull/882

cc @JunTaoLuo @rynowak @javiercn

### DIFF
--- a/src/Microsoft.AspNetCore.Antiforgery/AntiforgeryOptions.cs
+++ b/src/Microsoft.AspNetCore.Antiforgery/AntiforgeryOptions.cs
@@ -14,8 +14,13 @@ namespace Microsoft.AspNetCore.Antiforgery
         private const string AntiforgeryTokenFieldName = "__RequestVerificationToken";
         private const string AntiforgeryTokenHeaderName = "RequestVerificationToken";
 
-        private string _cookieName;
         private string _formFieldName = AntiforgeryTokenFieldName;
+
+        private CookieBuilder _cookieBuilder = new CookieBuilder
+        {
+            SameSite = SameSiteMode.Strict,
+            HttpOnly = true
+        };
 
         /// <summary>
         /// The default cookie prefix, which is ".AspNetCore.Antiforgery.".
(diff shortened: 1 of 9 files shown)


src/Microsoft.AspNetCore.Antiforgery/AntiforgeryOptions.cs
```diff
@@ -14,81 +14,78 @@ public class AntiforgeryOptions
...
+        [Obsolete("This property is obsolete and will be removed in a future version. The recommended alternative is " + nameof(Cookie) + "." + nameof(CookieBuilder.Domain) + ".")]
+        public string CookieDomain { get => Cookie.Domain; set => Cookie.Domain = value; }
 
         /// <summary>
-        /// Configures the <see cref="CookieOptions"/> of the antiforgery cookies. Without additional configuration, the 
-        /// default values antiforgery cookie options are true for <see cref="CookieOptions.HttpOnly"/>, null for
-        /// <see cref="CookieOptions.Domain"/> and <see cref="SameSiteMode.Strict"/> for <see cref="CookieOptions.SameSite"/>.
+        /// Determines the settings used to create the antiforgery cookies.
         /// </summary>
-        public Action<HttpContext, CookieOptions> ConfigureCookieOptions { get; set; }
```
rynowak: Does cookie builder have something similar to this?

rynowak:
:shipit: this looks great, I just want to double confirm that the builder has something like our configure delgate that's getting removed.

src/Microsoft.AspNetCore.Antiforgery/AntiforgeryOptions.cs
natemcmaster: No, but we made `CookieBuilder.Build(HttpContext context)` virtual and `AntiforgeryOptions.Cookie` settable for edge cases where users need to do something with HttpContext that we don't expect.

rynowak: We added this for a reason, so we'll have to figure out an alternative.

natemcmaster: What was the reason? When we were designing CookieBuilder, we figured anything other than the already-available top-level properties is an uncommon usage of this API.

src/Microsoft.AspNetCore.Antiforgery/AntiforgeryOptions.cs
```diff
@@ -111,5 +108,20 @@ public string FormFieldName
...
+            public AntiforgeryCookieBuilder()
+            {
+                SameSite = SameSiteMode.Strict;
+                HttpOnly = true;
+            }
+
+            public override string Name
+            {
+                get => base.Name;
+                set => base.Name = value ?? throw new ArgumentNullException(nameof(value));
```
Tratcher: null or empty?

natemcmaster: 1.x behavior only checks for null and only throws ArgumentNullException.

src/Microsoft.AspNetCore.Antiforgery/Internal/DefaultAntiforgeryTokenStore.cs
```diff
@@ -69,22 +69,14 @@ public void SaveCookieToken(HttpContext httpContext, string token)
...
-#pragma warning restore 618
-                SameSite = SameSiteMode.Strict,
-                Secure = _options.RequireSsl
-            };
-
-#pragma warning disable 618
-            if (_options.CookiePath != null)
+            var options = _options.Cookie.Build(httpContext);
+
+            options.Secure = options.Secure || _options.RequireSsl;
```
Tratcher: Why not depend on CookieSecurePolicy?

src/Microsoft.AspNetCore.Antiforgery/AntiforgeryOptions.cs
```diff
@@ -14,81 +14,78 @@ public class AntiforgeryOptions
...
-        public Action<HttpContext, CookieOptions> ConfigureCookieOptions { get; set; }
+        /// <remarks>
+        /// <para>
+        /// If an explicit <see cref="CookieBuilder.Name"/> is not provided, the system will automatically generate a
+        /// unique name that begins with <see cref="DefaultCookiePrefix"/>.
+        /// </para>
+        /// <para>
+        /// <see cref="CookieBuilder.SameSite"/> defaults to <see cref="SameSiteMode.Strict"/>.
+        /// <see cref="CookieBuilder.HttpOnly"/> defaults to <c>true</c>.
+        /// <see cref="CookieBuilder.SecurePolicy"/> will be forced to <seealso cref="CookieSecurePolicy.Always"/> when <see cref="RequireSsl"/> is <c>true</c>.
```
Tratcher: Map RequireSsl to Cookie.SecurePolicy?

src/Microsoft.AspNetCore.Antiforgery/AntiforgeryOptions.cs
rynowak: I know we did this to enable per-request configuration of things like the path, but I can't find an actual customer request for this, so I think we can just forget it until someone asks. Like you said, build is virtual, so we could use our subclass to resurrect this if necessary

src/Microsoft.AspNetCore.Antiforgery/Internal/DefaultAntiforgeryTokenStore.cs
rynowak: This has historically been here along with `RequireSsl`. I'm not really sure if the reasons are important or not.

rynowak:
:shipit: from me

src/Microsoft.AspNetCore.Antiforgery/Internal/DefaultAntiforgeryTokenStore.cs
natemcmaster: I looked through the code, and didn't find any significant reason for these to be separate options. I've pushed an update that obsoletes `RequireSsl` and forwards it to `Cookie.SecurePolicy`.

rynowak:
RequireSsl does more than just configure the cookie policy https://github.com/aspnet/Antiforgery/blob/f258be61fd05ff23d286661c58409b1eba004440/src/Microsoft.AspNetCore.Antiforgery/Internal/DefaultAntiforgery.cs#L268

src/Microsoft.AspNetCore.Antiforgery/Internal/DefaultAntiforgery.cs
```diff
@@ -265,12 +265,11 @@ private void SaveCookieTokenAndHeader(HttpContext httpContext, string cookieToke
 
         private void CheckSSLConfig(HttpContext context)
         {
-            if (_options.RequireSsl && !context.Request.IsHttps)
+            if (_options.Cookie.SecurePolicy == CookieSecurePolicy.Always && !context.Request.IsHttps)
```
natemcmaster: cc @rynowak

rynowak: RequireSsl does more than just configure the cookie policy https://github.com/aspnet/Antiforgery/blob/f258be61fd05ff23d286661c58409b1eba004440/src/Microsoft.AspNetCore.Antiforgery/Internal/DefaultAntiforgery.cs#L268

rynowak: Ok cool 👍

PR #150, opened 2017-07-03
Set "TreatWarningsAsErrors" before NuGet restore

mikeharder:
* Ensures our build stays clean of NuGet warnings


mikeharder:
I plan to make the same change in all repos at https://github.com/aspnet/Universe/blob/dev/build/Repositories.props.

PR #163, opened 2017-08-29
Sets the default cookie secure policy to None

kichalla:


src/Microsoft.AspNetCore.Antiforgery/AntiforgeryOptions.cs
```diff
@@ -19,7 +19,8 @@ public class AntiforgeryOptions
         private CookieBuilder _cookieBuilder = new CookieBuilder
         {
             SameSite = SameSiteMode.Strict,
-            HttpOnly = true
+            HttpOnly = true,
+            SecurePolicy = CookieSecurePolicy.None
```
pranavkm: Trailing comma.

rynowak: It would be good to have a comment here explaining the issues. We don't anyone auditing this code to see `Secure = None` and have alarms go off.

test/Microsoft.AspNetCore.Antiforgery.Test/Internal/DefaultAntiforgeryTokenStoreTest.cs
```diff
@@ -236,9 +236,16 @@ public void GetCookieToken_CookieIsNotEmpty_ReturnsToken()
...
-        [InlineData(CookieSecurePolicy.Always, true)]
-        [InlineData(CookieSecurePolicy.None, null)]
-        public void SaveCookieToken(CookieSecurePolicy policy, bool? expectedCookieSecureFlag)
+        [InlineData(false, CookieSecurePolicy.SameAsRequest, null)]
+        [InlineData(true, CookieSecurePolicy.SameAsRequest, true)]
+        [InlineData(false, CookieSecurePolicy.Always, true)]
+        [InlineData(true, CookieSecurePolicy.Always, true)]
+        [InlineData(false, CookieSecurePolicy.None, null)]
+        [InlineData(true, CookieSecurePolicy.None, null)]
+        public void SaveCookieToken_HonorsCookieSecurePolicy_OnOptions(
```
pranavkm: Could we split these out in to meaningful scenarios:

`SaveCookieToken_UsesRequestPolicy_WhenCookieSecurePolicyIsSameAsRequest`
`SaveCookieToken_CreatesSecurePolicyCookies_WhenCookieSecurePolicyIsAlways`
`SaveCookieToken_UsesValueFromConfig_WhenCookieSecurePolicyIsNone`

We can re-use the actual code as a private method but it'll be easier to understand what specific scenario is broken if this regresses.

src/Microsoft.AspNetCore.Antiforgery/AntiforgeryOptions.cs
kichalla: How does the following look?
```
Chrome recently deprecated the support of non-secure endpoints to set cookies with a 'secure' flag or overwrite cookies whose 'secure' flag is set (http://httpwg.org/http-extensions/draft-ietf-httpbis-cookie-alone.html). Since mixing secure and non-secure endpoints is a common scenario in applications, we are relaxing the restriction on secure policy on some cookies by setting to None. However, cookies related to authentication or authorization would still have restrictions as usual.
```
cc @Eilon

test/Microsoft.AspNetCore.Antiforgery.Test/Internal/DefaultAntiforgeryTokenStoreTest.cs
kichalla: Hmm, I think it looks fine as of now. It isn't difficult to follow even if it regresses.

src/Microsoft.AspNetCore.Antiforgery/AntiforgeryOptions.cs
rynowak: A few small edits:
```
Some browsers do not allow non-secure endpoints to set cookies with a 'secure' flag or overwrite cookies whose 'secure' flag is set (http://httpwg.org/http-extensions/draft-ietf-httpbis-cookie-alone.html). Since mixing secure and non-secure endpoints is a common scenario in applications, we are relaxing the restriction on secure policy on some cookies by setting to 'None'. Cookies related to authentication or authorization use a stronger policy than 'None'.
```

kichalla:
12e9430d89b85c1be0029ba2cded7840b2c59999

PR #177, opened 2018-01-09
Throwing AntiforgeryValidationException for failure to deserializing …

kichalla:
…tokens

Related to https://github.com/aspnet/Home/issues/2410 : Tampered antiforgery token should not result in InvalidOperationException
  
For reference: https://github.com/aspnet/Mvc/blob/dev/src/Microsoft.AspNetCore.Mvc.ViewFeatures/Internal/ValidateAntiforgeryTokenAuthorizationFilter.cs#L45

### DIFF
--- a/src/Microsoft.AspNetCore.Antiforgery/AntiforgeryValidationException.cs
+++ b/src/Microsoft.AspNetCore.Antiforgery/AntiforgeryValidationException.cs
@@ -12,11 +12,22 @@ namespace Microsoft.AspNetCore.Antiforgery
     {
         /// <summary>
         /// Creates a new instance of <see cref="AntiforgeryValidationException"/> with the specified
-        /// exception <paramref name="message"/>.
+        /// exception message.
         /// </summary>
         /// <param name="message">The message that describes the error.</param>
         public AntiforgeryValidationException(string message)
             : base(message)
+        {
+        }
+
+        /// <summary>
+        /// Creates a new instance of <see cref="AntiforgeryValidationException"/> with the specified
+        /// exception message and inner exception.
+        /// </summary>
+        /// <param name="message">The message that describes the error.</param>
+        /// <param name="innerException">The inner <see cref="Exception"/>.</param>
+        public AntiforgeryValidationException(string message, Exception innerException)
+            : base(message, innerException)
         {
         }
     }
--- a/src/Microsoft.AspNetCore.Antiforgery/Internal/DefaultAntiforgery.cs
+++ b/src/Microsoft.AspNetCore.Antiforgery/Internal/DefaultAntiforgery.cs
@@ -124,7 +124,10 @@ namespace Microsoft.AspNetCore.Antiforgery.Internal
             // Extract cookie & request tokens
             AntiforgeryToken deserializedCookieToken;
             AntiforgeryToken deserializedRequestToken;
-            DeserializeTokens(httpContext, tokens, out deserializedCookieToken, out deserializedRequestToken);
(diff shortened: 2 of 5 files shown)


src/Microsoft.AspNetCore.Antiforgery/Internal/DefaultAntiforgery.cs
```diff
@@ -197,11 +197,19 @@ private void ValidateTokens(HttpContext httpContext, AntiforgeryTokenSet antifor
...
+                DeserializeTokens(
+                    httpContext,
+                    antiforgeryTokenSet,
+                    out deserializedCookieToken,
+                    out deserializedRequestToken);
+            }
+            catch (InvalidOperationException ex)
+            {
+                throw new AntiforgeryValidationException(Resources.Antiforgery_FailedToDeserializeTokens, ex);
+            }
```
rynowak: I think this needs a slightly different solution, especially since we own the token serializer code. Putting a big try/catch with a general exception type like `IOE` around this much code seem sloppy 👎 I'd much rather see this inside the token serializer - which I believe is not extensible.

We also need to consider the other call to `DeserializeTokens` inside `TryValidate...` that needs to be updated as well.

kichalla (commented):
🆙 📅

src/Microsoft.AspNetCore.Antiforgery/AntiforgeryValidationException.cs
```diff
@@ -19,5 +19,16 @@ public AntiforgeryValidationException(string message)
             : base(message)
         {
         }
+
+        /// <summary>
+        /// Creates a new instance of <see cref="AntiforgeryValidationException"/> with the specified
+        /// exception <paramref name="message"/> and inner exception <paramref name="innerException"/>.
```
rynowak: In general you don't need to mix a paramref and a text description of the parameter. Remember that the parameter refers to the parameter, not the type of the parameter.

I'd just write this as `Creates a new instance of <see cref="AntiforgeryValidationException"/> with the specified exception message and inner exception`.

src/Microsoft.AspNetCore.Antiforgery/Internal/DefaultAntiforgery.cs
```diff
@@ -430,6 +434,27 @@ private AntiforgeryTokenSet Serialize(IAntiforgeryFeature antiforgeryFeature)
...
+            AntiforgeryTokenSet antiforgeryTokenSet,
+            out AntiforgeryToken cookieToken,
+            out AntiforgeryToken requestToken)
+        {
+            try
+            {
+                DeserializeTokens(httpContext, antiforgeryTokenSet, out cookieToken, out requestToken);
+                return true;
+            }
+            catch (Exception ex)
```
rynowak: I have the same feedback I have about the previous code. This is a try/catch that surrounds rather **a lot** of code. The try/catch needs to move down a layer into the serializer **or** this needs to catch specific exception type.

This has the same risks as the input formatter thing 😆

kichalla: Yeah, I pondered about this and used `Exception` as this is a `Try..` method, but as you mentioned (and I agree) that catching this root exception type is never good.

kichalla (commented):
🆙 📅

rynowak:
> that catching this root exception type is never good.

It's not that. There are use cases to catch `Exception`, but this isn't one of them.

rynowak (approved):
Much better

PR #178, opened 2018-01-10
Marked Antiforgery cookie as essential

kichalla:
Related to https://github.com/aspnet/Home/issues/2408 : Determine if the antiforgery cookie is considered essential

### DIFF
--- a/src/Microsoft.AspNetCore.Antiforgery/AntiforgeryOptions.cs
+++ b/src/Microsoft.AspNetCore.Antiforgery/AntiforgeryOptions.cs
@@ -21,6 +21,9 @@ namespace Microsoft.AspNetCore.Antiforgery
             SameSite = SameSiteMode.Strict,
             HttpOnly = true,
 
+            // Check the comment on CookieBuilder for more details
+            IsEssential = true,
+
             // Some browsers do not allow non-secure endpoints to set cookies with a 'secure' flag or overwrite cookies
             // whose 'secure' flag is set (http://httpwg.org/http-extensions/draft-ietf-httpbis-cookie-alone.html).
             // Since mixing secure and non-secure endpoints is a common scenario in applications, we are relaxing the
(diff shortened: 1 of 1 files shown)


src/Microsoft.AspNetCore.Antiforgery/AntiforgeryOptions.cs
```diff
@@ -20,6 +20,7 @@ public class AntiforgeryOptions
         {
             SameSite = SameSiteMode.Strict,
             HttpOnly = true,
+            IsEssential = true,
```
Tratcher: Comment that it's essential because otherwise form posts won't work?

Tratcher (approved):

src/Microsoft.AspNetCore.Antiforgery/AntiforgeryOptions.cs
```diff
@@ -45,6 +46,8 @@ public class AntiforgeryOptions
         /// <para>
         /// <see cref="CookieBuilder.SameSite"/> defaults to <see cref="SameSiteMode.Strict"/>.
         /// <see cref="CookieBuilder.HttpOnly"/> defaults to <c>true</c>.
+        /// <see cref="CookieBuilder.IsEssential"/> defaults to <c>true</c>.
```
rynowak: This block might be a good place to put an explanation.

kichalla (commented):
🆙 📅

src/Microsoft.AspNetCore.Antiforgery/AntiforgeryOptions.cs
```diff
@@ -45,6 +48,10 @@ public class AntiforgeryOptions
         /// <para>
         /// <see cref="CookieBuilder.SameSite"/> defaults to <see cref="SameSiteMode.Strict"/>.
         /// <see cref="CookieBuilder.HttpOnly"/> defaults to <c>true</c>.
+        /// <see cref="CookieBuilder.IsEssential"/> defaults to <c>true</c>. If this is marked as <c>false</c>, then
+        /// this cookie token is not sent in the response and any further requests to the server which require
+        /// antiforgery validation would fail due to this missing cookie token.
```
rynowak: This isn't really accurate.

I'd suggest something like:
``
The cookie used but the antiforgery system is part of a security system that is necessary when using cookie-based authentication. It should be consider required for the application to function.
```

PR #179, opened 2018-01-22
Updated the DefaultAntiforgery to set the the cache headers only if…

mkArtakMSFT:
… they aren't set yet.

This was resulting in an exception when a razor view was trying to render a form, when the headers have already been set.

### DIFF
--- a/src/Microsoft.AspNetCore.Antiforgery/Internal/DefaultAntiforgery.cs
+++ b/src/Microsoft.AspNetCore.Antiforgery/Internal/DefaultAntiforgery.cs
@@ -24,11 +24,11 @@ namespace Microsoft.AspNetCore.Antiforgery.Internal
         private readonly ILogger<DefaultAntiforgery> _logger;
 
         public DefaultAntiforgery(
-                    IOptions<AntiforgeryOptions> antiforgeryOptionsAccessor,
-                    IAntiforgeryTokenGenerator tokenGenerator,
-                    IAntiforgeryTokenSerializer tokenSerializer,
-                    IAntiforgeryTokenStore tokenStore,
-                    ILoggerFactory loggerFactory)
+            IOptions<AntiforgeryOptions> antiforgeryOptionsAccessor,
+            IAntiforgeryTokenGenerator tokenGenerator,
+            IAntiforgeryTokenSerializer tokenSerializer,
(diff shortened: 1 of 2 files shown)


mkArtakMSFT:
I'm now updating the tests to cover this case!

mkArtakMSFT:
Looks like the current design relies on "overriding" behavior. So I have to fix this, please don't review yet.

src/Microsoft.AspNetCore.Antiforgery/Internal/DefaultAntiforgery.cs
```diff
@@ -374,13 +374,28 @@ private IAntiforgeryFeature GetTokensInternal(HttpContext httpContext)
         /// <param name="httpContext">The <see cref="HttpContext"/>.</param>
         protected virtual void SetDoNotCacheHeaders(HttpContext httpContext)
         {
-            // Since antifogery token generation is not very obvious to the end users (ex: MVC's form tag generates them
-            // by default), log a warning to let users know of the change in behavior to any cache headers they might
-            // have set explicitly.
-            LogCacheHeaderOverrideWarning(httpContext.Response);
+            bool cacheHeadersChanged = SetHeaderIfNotSet(httpContext, HeaderNames.CacheControl, "no-cache, no-store");
```
javiercn: I don't think this will fix the problem, the reason is the cookie is also a header, which is the main reason why it will keep failing.

What we need to do is generate the cookie/form antiforgery components and compare them to any existing one (if any) on the current request. If they are the same, we don't do anything. If they are different, we throw (as the identity of the user changed after generating tokens for the first time).

javiercn:
Approach looks promising so far. Add some tests :)

test/Microsoft.AspNetCore.Antiforgery.Test/Internal/DefaultAntiforgeryTest.cs
```diff
@@ -1196,6 +1238,7 @@ public void GetAndStoreTokens_LogsWarning_NonNoCacheHeadersAlreadyPresent(string
             var antiforgery = GetAntiforgery(context);
             context.HttpContext.Response.Headers[headerName] = headerValue;
 
+            //context.HttpContext.Features
```
javiercn: Cleanup this?

PR #182, opened 2018-02-21
Sanitize build number

pranavkm:
Required to address https://github.com/aspnet/Coherence-Signed/issues/680

### DIFF
--- a/version.props
+++ b/version.props
@@ -5,7 +5,7 @@
     <PackageVersion Condition="'$(IsFinalBuild)' == 'true' AND '$(VersionSuffix)' == 'rtm' ">$(VersionPrefix)</PackageVersion>
     <PackageVersion Condition="'$(IsFinalBuild)' == 'true' AND '$(VersionSuffix)' != 'rtm' ">$(VersionPrefix)-$(VersionSuffix)-final</PackageVersion>
     <BuildNumber Condition="'$(BuildNumber)' == ''">t000</BuildNumber>
-    <SanitizedBuildNumber>$([System.Text.RegularExpressions.Regex]::Replace('$(BuildNumber)', '[^\w-]', '-'))</SanitizedBuildNumber>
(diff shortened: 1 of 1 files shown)


version.props
```diff
@@ -5,6 +5,7 @@
     <PackageVersion Condition="'$(IsFinalBuild)' == 'true' AND '$(VersionSuffix)' == 'rtm' ">$(VersionPrefix)</PackageVersion>
     <PackageVersion Condition="'$(IsFinalBuild)' == 'true' AND '$(VersionSuffix)' != 'rtm' ">$(VersionPrefix)-$(VersionSuffix)-final</PackageVersion>
     <BuildNumber Condition="'$(BuildNumber)' == ''">t000</BuildNumber>
-    <VersionSuffix Condition="'$(VersionSuffix)' != '' And '$(BuildNumber)' != ''">$(VersionSuffix)-$(BuildNumber)</VersionSuffix>
+    <SanitizedBuildNumber>$([System.Text.RegularExpressions.Regex]::Replace('$(BuildNumber)', '[^\w-]', '-'))</SanitizedBuildNumber>
```
natemcmaster: Any reason this shouldn't just replace the original value of BuildNumber instead of making a new property? BuildNumber is used in other places

pranavkm: This doesn't work no - `dotnet msbuild /p:BuildNumber=branch/name`

natemcmaster: So you are planning to pass `/p:BuildNumber=%teamcity.build.branch%/%teamcity.build.number%` on build? That might cause some problems because you would need to update every usage of `$(BuildNumber)` to `$(SanitizedBuildNumber)`. Maybe it would be easier to set a new property instead of overriding an existing one. For example, we could do

args: `/p:FeatureBranchVersionSuffix=%teamcity.build.branch%`
version.props:
```xml
<VersionSuffix>preview1</VersionSuffix>
<VersionSuffix Condition=" '$(p:FeatureBranchVersionSuffix)' != '' ">$(VersionSuffix)-$([System.Text.RegularExpressions.Regex]::Replace('$(p:FeatureBranchVersionSuffix)', '[^\w-]', '-'))</VersionSuffix>
```

pranavkm: Sure, that works

